PR #2, opened 2020-06-09
Adding password hashing functionality

kevinlewi:
Introduces a new trait, SlowHash, which is used by the client registration/login finish steps in order to run the slow hashing function after computing the OPRF key.

Also adds implementations for a NoOpHash and Scrypt (pulling in an scrypt dep.)


src/slow_hash.rs
```diff
@@ -0,0 +1,39 @@
...
+
+impl SlowHash for NoOpHash {
+    fn hash(
+        input: GenericArray<u8, <Sha256 as Digest>::OutputSize>,
+    ) -> Result<Vec<u8>, InternalPakeError> {
+        Ok(input.to_vec())
+    }
+}
+
+pub struct Scrypt;
```
huitseeker, line 28: you could do an `impl SlowHash` for `ScryptParams` instead, which would incidentally let anybody that knows how to construct a set of ScryptParams configure this.

kevinlewi:
Added slow-hash feature.

Yes, having examples of this would be helpful. Hopefully once we finish #4 the API will be simpler as well.

src/slow_hash.rs
```diff
@@ -0,0 +1,38 @@
...
+        Ok(input.to_vec())
+    }
+}
+
+#[cfg(feature = "slow-hash")]
+impl SlowHash for scrypt::ScryptParams {
+    fn hash(
+        input: GenericArray<u8, <Sha256 as Digest>::OutputSize>,
+    ) -> Result<Vec<u8>, InternalPakeError> {
+        let params = scrypt::ScryptParams::new(15, 8, 1).unwrap();
```
huitseeker, line 32: I was more thinking of using the `Self` parameter you get for any implementation. That way you get an 
```
fn hash(
        &self,
        input: GenericArray<u8, <Sha256 as Digest>::OutputSize>,
    ) -> Result<Vec<u8>, InternalPakeError> {
        let params = &self;
```

huitseeker: Never mind, I'm getting confused.

PR #6, opened 2020-06-12
Key cleanup: prove X25519KeyPair == SignalKeyPair, remove SignalKeyPair

huitseeker:
I advise looking at the second commit, which adds proptests showing those key types are precisely equal.

Since I remove those tests in the last commit (they operate on a SignalKeyPair that's removed in the present PR), they are only visible in the 2d commit, but they do pass there.


Cargo.toml
```diff
@@ -28,3 +28,5 @@ chacha20poly1305 = "0.4.1"
 hex = "0.4.2"
 lazy_static = "1.4.0"
 serde_json = "1.0"
+proptest = "0.10.0"
+rand = "0.7"
```
kevinlewi, line 32: newline

PR #8, opened 2020-06-15
Adding CipherSuite trait to handle bundling of underlying crypto primitives

kevinlewi:
In order to make the high-level API easier to use, I am adding the CipherSuite trait to allow the user to specify which primitives they want to across all of the ClientRegstration, ClientLogin, SeverRegistration, ServerLogin structs, rather than having to specify the primitives to each one individually.

### DIFF
--- a/src/ciphersuite.rs
+++ b/src/ciphersuite.rs
@@ -1,6 +1,30 @@
+// Copyright (c) Facebook, Inc. and its affiliates.
+//
+// This source code is licensed under the MIT license found in the
+// LICENSE file in the root directory of this source tree.
 
+//! Defines the CipherSuite trait to specify the underlying primitives for OPAQUE
+
+use crate::{
+    errors::InternalPakeError,
+    group::Group,
+    keypair::{Key, KeyPair},
+    slow_hash::SlowHash,
+};
+use generic_array::typenum::{U32, U64};
+use rand_core::{CryptoRng, RngCore};
+
+/// Configures the underlying primitives used in OPAQUE
 pub trait CipherSuite {
-    type Aead;
-    type Group;
-    type Keypair;
+    type Aead: aead::NewAead<KeySize = U32> + aead::Aead;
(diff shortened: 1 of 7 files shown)


src/ciphersuite.rs
```diff
@@ -0,0 +1,30 @@
...
+    errors::InternalPakeError,
+    group::Group,
+    keypair::{Key, KeyPair},
+    slow_hash::SlowHash,
+};
+use generic_array::typenum::{U32, U64};
+use rand_core::{CryptoRng, RngCore};
+
+/// Configures the underlying primitives used in OPAQUE
+pub trait CipherSuite {
```
huitseeker, line 18: So this is not bad, per se, but I still need to take a token struct `struct MyChoices()` and do an `impl for MyChoice`.

Whereas if you had a 
```
pub struct CipherSuite<Aead, Group, KeyFormat, SlowHash> 
where Aead: aead::NewAead<KeySize = U32> + aead::Aead, 
  Group: Group<ScalarLen = U32, UniformBytesLen = U64>, 
  ...
{
  aead: Phantom<Aead>,
  group: Phantom<Group>,
...
}

impl<Aead, Group, KeyFormat, SlowHash> CypherSuite<Aead, Group, KeyFormat, SlowHash> {

fn generate_random_keypair<R: RngCore + CryptoRng>(
        rng: &mut R,
    ) -> Result<KeyFormat, InternalPakeError> {
        <KeyFormat as KeyPair>::generate_random(rng)
    }
}
```

Then the instantiation becomes easier, especially since we have already implemented the smaller traits in this lib.

```
type MyChoices = CipherSuite<AesGcm, EdwardsPoint, X25519KeyPair, NoOpHash>;
```

kevinlewi: After some offline discussion, we decided to keep the trait syntax as it might be cleaner for the end user

kevinlewi:
Created #10 as a follow-up

PR #13, opened 2020-06-22
Updated CONTRIBUTING.md

zer0x64:
I removed reference to Facebook and Ristretto255.js, since I don't think it doesn't really apply here.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -26,5 +26,5 @@ disclosure of security bugs. In those cases, please go through the process
 outlined on that page and do not file a public issue.
 
 ## License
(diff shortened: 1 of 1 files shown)


kevinlewi:
Thanks for the Ristretto255.js -> opaque-ke catch!

However, the Facebook statements regarding the CLA here still apply, since Novi is a subsidiary of Facebook. cc: @ericnakagawa in case I am wrong about this.

zer0x64:
Didn't know about it! I updated the PR to re-add the Facebook-related parts

zer0x64:
I also submitted the CLA, although I'm not sure how useful it is since I'm technically not contributing code here

PR #14, opened 2020-06-23
Updating to draft-krawczyk-cfrg-opaque-06, reworking envelope construction and removing AEAD

kevinlewi:
In https://tools.ietf.org/html/draft-krawczyk-cfrg-opaque-06#section-4, a new procedure for generating the envelope has now been specified: rather than relying on AEAD encryption that is RKR-secure, we now use an XOR-based mechanism instead.

This means that the CipherSuite object no longer needs to specify an AEAD.

I also renamed rkr_encryption to envelope.


src/envelope.rs
```diff
@@ -0,0 +1,173 @@
...
+use hmac::{Hmac, Mac, NewMac};
+use rand_core::{CryptoRng, RngCore};
+use sha2::{Digest, Sha256};
+
+// Constant string used as salt for HKDF computation
+const STR_ENVU: &[u8] = b"EnvU";
+
+/// The length of the "export key" output by the client registration
+/// and login finish steps
+pub(crate) type ExportKeySize = <Sha256 as Digest>::OutputSize;
```
huitseeker, line 21: (for my edification:) Do we plan the digest to change, here? In particular I'm wondering w.r.t #15 if there are areas where a change in digests makes more or less sense than others.

src/envelope.rs
```diff
@@ -0,0 +1,173 @@
...
+
+/// This struct is a straightforward instantiation of the trait separating the
+/// three components in Vecs
```
huitseeker, line 24: I think we need a docstring that's far less "internal dev notes" and more specific in its reference to the spec. In particular, here, we are taking the "storage/communication efficient EnvU" option described in the spec in §4, and should say so IMHO. Moreover, there's no more trait :D

src/envelope.rs
```diff
@@ -0,0 +1,173 @@
...
+    /// The additional number of bytes added to the plaintext
+    pub(crate) fn additional_size() -> usize {
+        Self::nonce_size() + <Sha256 as Digest>::OutputSize::to_usize()
+    }
+
+    fn hmac_key_size() -> usize {
+        <Sha256 as Digest>::OutputSize::to_usize()
+    }
+
+    fn hmac_size() -> usize {
```
huitseeker, line 43: I suspect we can replace this (used only in `from_bytes`) with `hmac_key_size`

src/envelope.rs
```diff
@@ -0,0 +1,173 @@
...
+            bytes[ciphertext_start..ciphertext_end].to_vec(),
+            GenericArray::from_slice(&bytes[ciphertext_end..]),
+        ))
+    }
+
+    pub(crate) fn to_bytes(&self) -> Vec<u8> {
+        [&self.nonce[..], &self.ciphertext[..], &self.hmac[..]].concat()
+    }
+
+    pub(crate) fn seal<R: RngCore + CryptoRng>(
```
huitseeker, line 85: Could we add a docstring on this one, especially one documenting the use of a fresh random nonce for each EnvU is necessary for the security of the enveloping scheme?

src/envelope.rs
```diff
@@ -0,0 +1,173 @@
...
+        let h = Hkdf::<Sha256>::new(Some(&nonce), &key);
+        let mut okm = vec![0u8; plaintext.len() + Self::hmac_key_size() + Self::export_key_size()];
+        h.expand(STR_ENVU, &mut okm)
+            .map_err(|_| InternalPakeError::HkdfError)?;
+        let xor_key = &okm[..plaintext.len()];
+        let hmac_key = &okm[plaintext.len()..plaintext.len() + Self::hmac_key_size()];
+        let export_key = &okm[plaintext.len() + Self::hmac_key_size()..];
+
+        let ciphertext: Vec<u8> = xor_key
+            .to_vec()
```
huitseeker, line 103: I suspect this `to_vec` may be superfluous

src/envelope.rs
```diff
@@ -0,0 +1,173 @@
...
+        let xor_key = &okm[..self.ciphertext.len()];
+        let hmac_key = &okm[self.ciphertext.len()..self.ciphertext.len() + Self::hmac_key_size()];
+        let export_key = &okm[self.ciphertext.len() + Self::hmac_key_size()..];
+
+        let mut hmac =
+            Hmac::<Sha256>::new_varkey(&hmac_key).map_err(|_| InternalPakeError::HmacError)?;
+        hmac.update(&self.nonce);
+        hmac.update(&self.ciphertext);
+        hmac.update(aad);
+        if hmac.verify(&self.hmac).is_err() {
```
huitseeker, line 140: we may want to be explicit —here or in the `thiserror`-powered message for `OpenHmacError`— about the issue revealed by a failure at this step. At a minimum, I think we want to say "opening worked, but the Hmac didnt" rather than what we say now, which is just the later.

src/opaque.rs
```diff
@@ -658,11 +629,7 @@ impl<CS: CipherSuite> ClientLogin<CS> {
     }
 }
 
-type ClientLoginFinishResult = (
-    LoginThirdMessage,
-    Vec<u8>,
-    GenericArray<u8, <Sha256 as Digest>::OutputSize>,
-);
+type ClientLoginFinishResult = (LoginThirdMessage, Vec<u8>, GenericArray<u8, ExportKeySize>);
```
huitseeker, line 632: Not that this is necessarily part of your PR, but I'm repetitively struggling at remembering the nature of the components of this tuple, something a proper struct with labels (third_message, shared_secret, export-key) would help with.

src/opaque.rs
```diff
@@ -880,6 +833,7 @@ impl ServerLogin {
         )?;
 
         let l2 = LoginSecondMessage {
+            _key_format: PhantomData,
```
huitseeker, line 836: :D

src/envelope.rs
kevinlewi: Yes, the digest here could change. I don't think we should hardcode Sha256 anywhere.

src/envelope.rs
kevinlewi: Technically the hmac key and the hmac itself can be of different sizes. I will leave it as-is for now unless you disagree!

src/envelope.rs
kevinlewi: Opening can no longer fail (because it is an XOR operation instead of a decryption operation now)

src/opaque.rs
```diff
@@ -762,35 +726,27 @@ impl<CS: CipherSuite> ClientLogin<CS> {
...
-                    PakeError::DecryptionHmacError => PakeError::InvalidLoginError,
-                    err => err,
-                })?,
-        )?;
+
+        let (client_s_sk, export_key) = &l2
+            .envelope
+            .open(&password_derived_key, &server_s_pk.to_arr())
+            .map_err(|e| match e {
+                PakeError::SealOpenHmacError => PakeError::InvalidLoginError,
```
huitseeker, line 734: nit: this makes me think `SealOpenHmacError` should be an `InternalPakeError` which would translate to `InvalidLoginError` at the `PakeError` level

src/tests/serialization.rs
```diff
@@ -67,16 +65,18 @@ fn server_registration_roundtrip() {
     assert_eq!(reg_bytes, oprf_bytes);
     // If we do have envelope and client pk, the server registration contains
     // the whole kit
-    let rkr_size = RKRCiphertext::<ChaCha20Poly1305>::rkr_with_nonce_size();
-    let mut mock_rkr_bytes = vec![0u8; rkr_size];
-    rng.fill_bytes(&mut mock_rkr_bytes);
-    println!("{}", mock_rkr_bytes.len());
+    let key_len =
+        <<<Default as CipherSuite>::KeyFormat as KeyPair>::Repr as SizedBytes>::Len::to_usize();
```
huitseeker, line 69: not: I suspect you do not need all the `as Trait` here, and there's a higher probability that the inner one is not required.

src/envelope.rs
huitseeker: Fair enough!

src/opaque.rs
huitseeker: Still there, but it's a nit

src/opaque.rs
kevinlewi: Converting all errors thrown from envelope.rs to be InternalPakeError instead of PakeError

src/tests/serialization.rs
kevinlewi: I tried removing each "as Trait", and each one causes the same error to appear.

src/opaque.rs
kevinlewi: Since this results in a change to the API, I will leave it to a future discussion where we consider the readability of the user-facing API as a whole

PR #18, opened 2020-06-29
Restructure hash to curve

kevinlewi:
Putting the hkdf evaluation within EdwardsPoint and changing the interface of the hash_to_curve function

### DIFF
--- a/src/group.rs
+++ b/src/group.rs
@@ -10,14 +10,10 @@ use curve25519_dalek::{
     ristretto::{CompressedRistretto, RistrettoPoint},
     scalar::Scalar,
 };
-use generic_array::{
-    typenum::{U32, U64},
-    ArrayLength, GenericArray,
-};
+use generic_array::{typenum::U32, ArrayLength, GenericArray};
+use hkdf::Hkdf;
 use rand_core::{CryptoRng, RngCore};
-
-use sha2::{Digest, Sha256};
-
+use sha2::{Digest, Sha256, Sha512};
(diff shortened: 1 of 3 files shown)


src/group.rs
```diff
@@ -52,8 +48,7 @@ pub trait Group: Sized + for<'a> Mul<&'a <Self as Group>::Scalar, Output = Self>
     /// impl is allowed to perform additional hashes if it needs to, but this
     /// may not be necessary as this function is going to be called with the
     /// output of a kdf.
-    type UniformBytesLen: ArrayLength<u8>;
-    fn hash_to_curve(uniform_bytes: &GenericArray<u8, Self::UniformBytesLen>) -> Self;
+    fn hash_to_curve(input: &[u8], pepper: Option<&[u8]>) -> Self;
```
huitseeker, line 51: Right, so as discussed online, the best approach here is to have a well-defined notion of hashing to the curve, either of a non-random byte slice representing your payload (i.e. you don't need the `UniformBytesLength`) or of uniform bytes (hence you need the lenght, because algos such as elligator2 have a strong requirement on the amount of entropy). Then you can have a default implementation for another function `password_to_curve` which calls into `hash_to_curve`, which you may want to split into an extension trait of `Group` if you expect people to overwrite the hkdf portion but not the group (which seems probable).

huitseeker: #21

kevinlewi:
Closing in favor of #21

PR #20, opened 2020-07-03
Introducing a trait for key exchange

kevinlewi:


src/key_exchange/traits.rs
```diff
@@ -0,0 +1,55 @@
...
+
+use crate::{
+    errors::{InternalPakeError, ProtocolError},
+    keypair::{Key, KeyPair},
+};
+use rand_core::{CryptoRng, RngCore};
+
+use std::convert::TryFrom;
+
+pub trait KeyExchange {
```
huitseeker, line 14: With a monolithic trait:
- the implementer has to implement both the Client and the Server part of the protocol monolithically. As we go through the work of having a binary format for messages, it opens an opportunity for having a client / server part independent of its counterpart.
- the implementer has to implement all messages in one fell swoop. Yet e.g. a protocol without the KE3 message (i.e. server and client meet on a shared-secret-derived channel rather than HMac), would kinda make sense.

src/key_exchange/tripledh.rs
```diff
@@ -0,0 +1,415 @@
...
+        KE1_STATE_LEN
+    }
+
+    fn ke2_message_size() -> usize {
+        KE2_MESSAGE_LEN
+    }
+}
+
+#[derive(PartialEq, Eq)]
+pub struct KE1State {
```
huitseeker, line 215: This could be `KE1State<KeyFormat: KeyPair>`, right? as it stands, couldn't we do XEdDSA (rather than X25519) for this KE? The use of that sort of thing is the integration with an existing protocol supported by a keypair of a different format:
> The integrated protocol will have as its first
> message the TLS's Client Hello augmented with user account
> information and with the DH-OPRF first message (the value alpha).
> The server's response includes the regular TLS 1.3 second flight
> augmented with the second OPRF message which includes the values beta
> and EnvU.  For its TLS signature, the server uses the private key
> PrivS whose corresponding public key PubS is authenticated as part of
> the user envelope EnvU (there is no need to send a regular TLS
> certificate in this case).

src/key_exchange/tripledh.rs
```diff
@@ -0,0 +1,415 @@
...
+    client_e_sk: Key,
```
huitseeker, line 216: https://tools.ietf.org/html/draft-krawczyk-cfrg-opaque-06#section-6
Does that make sense?

src/key_exchange/traits.rs
kevinlewi: Created https://github.com/novifinancial/opaque-ke/issues/23 to track this, but let's push this first since this change might not be the highest priority at the moment.

src/key_exchange/tripledh.rs
kevinlewi: Added a note for this in https://github.com/novifinancial/opaque-ke/issues/23

Issue #15, opened 2020-06-23
Add the ability to specify the underlying hash in the CipherSuite object


kevinlewi:
Right now, there are bunch of places in the code where Sha256 is hard-coded -- we should make this customizable.

kevinlewi:
This has been completed.

PR #21, opened 2020-07-03
Restructure password-hashing-to-the-curve as an extension trait of Group

huitseeker:
This takes the main insight of #18 and splits it into the `Group` component, on the one hand, and the password hashing, on the other hand.

As discussed offline those are essentially the two components of password-to-curve-point hashing:
- password-to-entropy-bytes,
- entropy-bytes-to-curve (which is essentially the vanilla `hash_to_curve` op of the underlying `Group`),

The first commit is trying to hint at what the code would look like if we wanted to offer some `Digest` flexibility in the password-to-entropy component.

THe second commit hints that the uses of `Sha256` we want to customize in #15 are actually *not* the length of the digest used to fit a quota of entropy bytes for the hash-to-curve operation. Rather we want to choose whether we want to use an HKDF at all. 

Hence the `Digest` parameter collapses into the impl choice within the `PasswordToCurve`, and if one wants customization, that's where it would be found (in a simpler fashion than yet-another `CipherSuite` parameter).

Supersedes #18

### DIFF
--- a/src/ciphersuite.rs
+++ b/src/ciphersuite.rs
@@ -7,24 +7,22 @@
 
 use crate::{
     errors::InternalPakeError,
-    group::Group,
     keypair::{Key, KeyPair},
-    oprf::HkdfDigest,
+    map_to_curve::GroupWithMapToCurve,
     slow_hash::SlowHash,
 };
-use digest::FixedOutput;
-use generic_array::typenum::{U32};
+
 use rand_core::{CryptoRng, RngCore};
 
 /// Configures the underlying primitives used in OPAQUE
-/// * `Digest`: a digest suitable for use in an Hkdf, with an output length equal
-///     to the input of the hash-to-curve function of the `Group` parameter.
-/// * `Group`: a finite cyclic group along with a point representation
+/// * `Group`: a finite cyclic group along with a point representation, along
+///   with an extension trait PasswordToCurve that allows some customization on
+///   how to hash a password to a curve point. See `group::Group` and
+///   `map_to_curve::GroupWithMapToCurve`.
 /// * `KeyFormat`: a keypair type composed of public and private components
 /// * `SlowHash`: a slow hashing function, typically used for password hashing
 pub trait CipherSuite {
-    type Digest: HkdfDigest;
-    type Group: Group<ScalarLen = U32, UniformBytesLen = <Self::Digest as FixedOutput>::OutputSize>;
+    type Group: GroupWithMapToCurve;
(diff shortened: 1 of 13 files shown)


src/ciphersuite.rs
```diff
@@ -7,19 +7,22 @@
...
-/// * SlowHash: a slow hashing function, typically used for password hashing
+/// * `Group`: a finite cyclic group along with a point representation, along
+///   with an extension trait PasswordToCurve that allows some customization on
+///   how to hash a password to a curve point. See `group::Group` and
+///   `password_to_curve::PasswordToCurve`.
+/// * `KeyFormat`: a keypair type composed of public and private components
+/// * `SlowHash`: a slow hashing function, typically used for password hashing
 pub trait CipherSuite {
-    type Group: Group<ScalarLen = U32, UniformBytesLen = U64>;
+    type Group: PasswordToCurve;
```
kevinlewi, line 25: Naming thing: Can we keep the top-level trait that is used in CipherSuite still called "Group", instead of PasswordToCurve? It does more than just turn a password into a curve point, it also does the group operations. Also, this should be kept more or less in sync with the lib.rs docs which will just refer to this as a group.

src/password_to_curve.rs
```diff
@@ -0,0 +1,23 @@
+use crate::group::Group;
+use curve25519_dalek::{edwards::EdwardsPoint, ristretto::RistrettoPoint};
+
+use generic_array::GenericArray;
+use hkdf::Hkdf;
+
+pub trait PasswordToCurve: Group {
```
kevinlewi, line 7: How about calling this MapToCurve instead of PasswordToCurve, at least?

And then how about doing:

pub trait GroupWithMapToCurve: Group { ... } ?

src/ciphersuite.rs
```diff
@@ -7,19 +7,22 @@
...
 use rand_core::{CryptoRng, RngCore};
 
 /// Configures the underlying primitives used in OPAQUE
-/// * Group: a finite cyclic group along with a point representation
-/// * KeyFormat: a keypair type composed of public and private components
-/// * SlowHash: a slow hashing function, typically used for password hashing
+/// * `Group`: a finite cyclic group along with a point representation, along
+///   with an extension trait PasswordToCurve that allows some customization on
+///   how to hash a password to a curve point. See `group::Group` and
+///   `password_to_curve::PasswordToCurve`.
```
kevinlewi, line 21: Comment needs updating to not reference PasswordToCurve anymore

PR #24, opened 2020-07-21
Adding a hash type to CipherSuite

kevinlewi:
- Removes sha2 as a dependency (moved to dev-dependencies)
- Introduces a Hash and a CurveHash associated type for the CipherSuite object

- Remaining item to address: Specification of the CurveHash should be tied to the choice of Group (UniformBytesLen should match OutputSIze)


src/digest.rs
```diff
@@ -0,0 +1,20 @@
...
+pub trait Hash: Update + BlockInput + FixedOutput + Reset + Default + Clone {
+    /// The block size for the hash function
+    type BlockSize: ArrayLength<u8>;
+    /// The output size of the hash function
+    type OutputSize: ArrayLength<u8>;
+}
+
+impl<T: Update + BlockInput + FixedOutput + Reset + Default + Clone> Hash for T {
+    type BlockSize = T::BlockSize;
+    type OutputSize = T::OutputSize;
```
huitseeker, line 19: I wish associated type defaults where stabilized. Since it's on its way, perhaps link to https://github.com/rust-lang/rust/issues/29661?
Thanks for the tip, btw, I should do just that in my SizedBytes PR.

src/group.rs
```diff
@@ -97,12 +100,11 @@ impl Group for RistrettoPoint {
...
     type UniformBytesLen = U64;
-    fn hash_to_curve(uniform_bytes: &GenericArray<u8, Self::UniformBytesLen>) -> Self {
-        // This is because RistrettoPoint is on an obsolete sha2 version, see https://github.com/dalek-cryptography/curve25519-dalek/pull/327
+    fn hash_to_curve<H: Hash>(
+        uniform_bytes: &GenericArray<u8, Self::UniformBytesLen>,
+    ) -> Self {
         let mut bits = [0u8; 64];
-        let mut hasher = sha2::Sha512::new();
-        hasher.update(uniform_bytes);
-        bits.copy_from_slice(&hasher.finalize());
```
huitseeker, line 105: Aren't we missing the actual hashing here? The point of the above three lines was to deterministically reproduce the results of [`Scalar::hash_from_bytes`](https://github.com/dalek-cryptography/curve25519-dalek/blob/c4824e1384ef188eb9284ff866e5ad6b814400f7/src/scalar.rs#L587-L593)

src/oprf.rs
```diff
@@ -139,16 +148,13 @@ mod tests {
...
+        let res = generate_oprf3::<RistrettoPoint, sha2::Sha256>(&input, alpha, &blinding_factor)
+            .unwrap();
 
         let (hashed_input, _) = Hkdf::<Sha512>::extract(None, &input);
-
-        // This is because RistrettoPoint is on an obsolete sha2 version
         let mut bits = [0u8; 64];
-        let mut hasher = sha2::Sha512::new();
-        Digest::update(&mut hasher, &hashed_input[..]);
-        bits.copy_from_slice(&hasher.finalize());
```
huitseeker, line 151: We seem to be missing the actual hashing here. See my comment above on Group.

src/group.rs
kevinlewi: Discussed offline -- this is not necessary since we are already hashing before passing into this function

kevinlewi:
Removed CurveHash

PR #25, opened 2020-07-21
[Serialization] Derive Serializable + use throughout KeyExchange & associated

huitseeker:
Up for early review. Builds on #29 which should get an early merge.

- `#[derive(Serializable)]` is now a thing (except for Unions/ Enums),
- used throughout `key_exchange`, not yet through opaque.rs


opaque-derive/Cargo.toml
```diff
@@ -0,0 +1,19 @@
+[package]
+name = "opaque-derive"
+version = "0.1.0"
+repository = "https://github.com/novifinancial/opaque-ke"
+authors = ["Kevin Lewi <klewi@fb.com>", "François Garillot <fga@fb.com>"]
+license = "MIT"
+edition = "2018"
```
kevinlewi, line 7: Do we need all of these here?

huitseeker: Well, now that we do have to publish opaque-derive, I'm afraid we do.

src/keypair.rs
```diff
@@ -95,36 +102,6 @@ trait KeyPairExt: KeyPair + Debug {
...
-
-    fn from_arr(arr: &GenericArray<u8, Self::Len>) -> Result<Self, InternalPakeError> {
-        let (public_key, private_key): (GenericArray<u8, T::Len>, GenericArray<u8, _>) =
-            GenericArray::split(arr.clone());
-        let public = <T as SizedBytes>::from_arr(&public_key)?;
-        let private = <T as SizedBytes>::from_arr(&private_key)?;
-        KP::new(public, private)
-    }
-}
-
```
huitseeker: @kevinlewi  This part is a really nice removal I think, replaced by the 1-line derivation of `SizedBytes` on `X25519KeyPair` on line 131.

huitseeker:
Closing in favor of https://github.com/novifinancial/generic-array-derive

PR #29, opened 2020-07-23
Maintenance

huitseeker:
Straightforward cleanup / updates.


Cargo.toml
```diff
@@ -17,25 +17,25 @@ u64_backend = ["curve25519-dalek/u64_backend", "x25519-dalek/u64_backend"]
 u32_backend = ["curve25519-dalek/u32_backend", "x25519-dalek/u32_backend"]
 
 [dependencies]
-aead = "0.3.1"
+aead = "0.3.2"
```
kevinlewi, line 20: You can actually just remove this aead dependency, since we do not depend on it anymore.

PR #35, opened 2020-08-05
Updating serialization for RegisterFirstMessage

kevinlewi:
Updates the RegisterFirstMessage struct to have a serialize / deserialize function which incorporates the TLS formatting according to the OPAQUE spec


src/opaque.rs
```diff
@@ -33,19 +36,48 @@ pub struct RegisterFirstMessage<Grp> {
 
 impl<Grp: Group> TryFrom<&[u8]> for RegisterFirstMessage<Grp> {
     type Error = ProtocolError;
-    fn try_from(first_message_bytes: &[u8]) -> Result<Self, Self::Error> {
+    fn try_from(input: &[u8]) -> Result<Self, Self::Error> {
+        if input[0] != REGISTRATION_REQUEST {
+            return Err(ProtocolError::VerificationError(
+                PakeError::SerializationError,
```
huitseeker, line 42: what about using an `into()` or `?` on `SerializationError` directly?

kevinlewi: Can you clarify what you mean by this?

PR #36, opened 2020-08-19
Check for small subgroup components when using the EdwardsPoint group for the OPRF

### LINKED ISSUE #34: Check for and eliminate mixed-order points in the Curve25519 implementation of OPRF
From the OPRF draft:
https://tools.ietf.org/html/draft-irtf-cfrg-voprf-04#section-2.1

   It is convenient in cryptographic applications to instantiate such
   prime-order groups using elliptic curves, such as those detailed in
   [SEC2].  For some choices of elliptic curves (e.g. those detailed in
   [RFC7748] require accounting for cofactors) there are some
   implementation issues that introduce inherent discrepancies between
   standard prime-order groups and the elliptic curve instantiation.  In
   this document, all algorithms that we detail assume that the group is
   a prime-order group, and this MUST be upheld by any implementer.
   That is, any curve instantiation should be written such that any
   discrepancies with a prime-order group instantiation are removed.
   See Section 4 for advice corresponding to implementation of this
   interface for specific definitions of elliptic curves.

=> Even though the consequences of a user providing a low or mixed order point to a server¹ are unclear to me, our "larger subgroup" curve25519 implementation of group just relies on the fact its provided generator has a codomain that's torsion-free, but does not check a user's submission is torsion-free

¹: they do reveal their pw in the low order case

huitseeker:
A malicious attacker sending a small point could lead the server / user to leak private information.

The check avoiding mixed order points (rather than just small) is prohibitively expensive, and I don't know how an attacker would extract any data from that => we focus on a small order check.

Fixes #34.


src/group.rs
```diff
@@ -164,3 +169,62 @@ impl Group for EdwardsPoint {
...
+        let mut bytes = [0u8; 32];
+        bytes.copy_from_slice(&pt[..32]);
+
+        curve25519_dalek::edwards::CompressedEdwardsY(bytes)
+            .decompress()
+            .ok_or_else(|| anyhow!("Point decompression failed!"))
+    }
+
+    #[test]
+    fn test_small_subgroup_Edwards() {
```
kevinlewi, line 223: This should be fixed

kevinlewi:
LGTM modulo the comment

PR #44, opened 2020-08-31
Fix omission of nonce in hmac computation for envelope

kevinlewi:
Hmac computation wasn't accounting for the nonce -- this was a bug

Also renaming opaque_key to export_key


kevinlewi:
Well, this would have been caught by matching against official test vectors, once they are published.

PR #45, opened 2020-09-02
Proptests exercising `try_from` for most structures based on random data

huitseeker:
This also fixes crashes in:
- RegisterFirstMessage,
- LoginFirstMessage,
- ClientRegistration,
- ClientLogin

I also think, retrospectively, that this demonstrates a need for a systematic approach to (de)serialization (which of course may or may not be generic-array-derive).


kevinlewi (approved):
Excellent!

PR #46, opened 2020-09-03
Implement and use Elligator2 for the Curve25519 larger subgroup instance

huitseeker:
The implementation is largely replicating
https://github.com/signalapp/libsignal-protocol-c/blob/master/src/curve25519/ed25519/additions/elligator.c

and of course, it resets the downstream test vectors.

This requires pulling some of the Field implementation from curve25519, which is not public. I've tried to minimize the imports using fiat-crypto.


kevinlewi (approved):
Thank you!

PR #49, opened 2020-09-10
Rebase TLS on version 0.2

huitseeker:
Let's use this as a review branch, even though I may have to make a manual push.


kevinlewi:
Thanks!

PR #50, opened 2020-09-19
Simplifications and Normalizations

huitseeker:
This PR rationalizes some of our (complex) uses of generics and types:
- TripleDH converts back-and-forth to and from `GenericArray` both on storage (in the `KE[1-4](State|Message)` types) and on entering / exiting the KE (`(to|from)_bytes`). The later is justified, but the former is not and this PR corrects it.
- We transform the hidden panics from `foo.copy_from_slice(x)` where `foo` is a `[u8; N]` into explicit `TryInto` instances (followed by expect + a justification).
- `LoginSecondMessage` now uses the `CipherSuite` instance rather than 4 different types.
- the `KeyExchange` trait now uses a single `KeyFormat` parameter for its 3 functions (just like `Hash`) instead of 3 generic functions (for which having distinct instantiations would make no sense).
- `Hash` defined two parameters `OutputSize` and `BlockSize`, the later unused, and the former without relationship with the one defined by the required supertrait of `FixedOutput` (hence introducing errors), we eliminate them.
- `sized_bytes_using_constant_and_try_from` was unused and is removed.

### DIFF
--- a/src/ciphersuite.rs
+++ b/src/ciphersuite.rs
@@ -34,7 +34,7 @@ pub trait CipherSuite {
     /// A keypair type composed of public and private components
     type KeyFormat: KeyPair<Repr = Key> + PartialEq;
     /// A key exchange protocol
-    type KeyExchange: KeyExchange<Self::Hash>;
+    type KeyExchange: KeyExchange<Self::Hash, Self::KeyFormat>;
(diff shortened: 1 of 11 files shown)


kevinlewi (approved):
Thank you!!

Issue #55, opened 2020-10-05
Should the client verify consistency of user-supplied idU and idS against Envelope?


kevinlewi:
At the moment, idU and idS are user-supplied, but not checked against the envelope that the server returns during the login phase

kevinlewi:
No longer relevant

PR #56, opened 2020-10-05
Incorporate idU, idS and support EnvelopeCredentialsFormat

kevinlewi:
This PR accomplishes two main things that are a part of [the spec](https://tools.ietf.org/html/draft-irtf-cfrg-opaque-00):

1) Supporting the client to supply an idU and idS that is attached to the RegisterFirstMessage / LoginFirstMessage payloads, and is also incorporated into how the Envelope is generated
2) Supporting a way for the server to specify the structure of the envelope using EnvelopeCredentialsFormat

For 1), this involved a change to the serialization of each of the serialization methods for each of these structs. I also improved the testing of the serialization to be more modular.

For 2), this meant a modification to the seal() and open() interfaces in Envelope to now take the EnvelopeCredentialsFormat for seal, rather than asking for a ciphertext/plaintext and aad. #55 is an issue that was created in relation to this change


src/envelope.rs
```diff
@@ -81,48 +125,99 @@ impl<D: Hash> Envelope<D> {
...
+        let mut ret = Vec::new();
+        for index_type in cred_format {
+            let byte: u8 = index_type as u8 + 1;
+            let chunk = match &credentials.get(&index_type) {
+                Some(v) => [vec![byte], serialize(v.to_vec(), 2)].concat(),
+                None => return Err(InternalPakeError::IncompatibleEnvelopeCredentialsError),
+            };
+            ret.extend(chunk);
+        }
+        Ok(ret)
```
huitseeker, line 185:
```suggestion
        let mut ret = Vec::new();
        for index_type in cred_format {
            match &credentials.get(&index_type) {
                Some(v) => {
                    ret.push(index_type as u8 + 1);
                    ret.extend(serialize(v.to_vec(), 2));
                }
                None => return Err(InternalPakeError::IncompatibleEnvelopeCredentialsError),
            }
        }
        Ok(ret)
```

Also `serialization::serialize` should clearly be fixed to take its argument by reference.

src/envelope.rs
```diff
@@ -81,48 +125,99 @@ impl<D: Hash> Envelope<D> {
...
+        bytes: &[u8],
+    ) -> Result<HashMap<CredentialType, Vec<u8>>, InternalPakeError> {
+        let mut credentials: HashMap<CredentialType, Vec<u8>> = HashMap::new();
+        let mut bytes_copy: Vec<u8> = Vec::new();
+        bytes_copy.extend_from_slice(&bytes);
+        while !bytes_copy.is_empty() {
+            let t = match u8_to_credential_type(bytes_copy[0]) {
+                Some(v) => v,
+                None => return Err(InternalPakeError::IncompatibleEnvelopeCredentialsError),
+            };
```
huitseeker, line 198:
```suggestion
            let t = u8_to_credential_type(bytes_copy[0])
                .ok_or_else(InternalPakeError::IncompatibleEnvelopeCredentialsError)?;
```

src/opaque.rs
```diff
@@ -154,18 +164,45 @@ where
...
+            .iter()
+            .map(|&x| u8_to_credential_type(x))
+            .collect();
+        let ecf = match sc {
+            Some(secret) => match cc {
+                Some(cleartext) => EnvelopeCredentialsFormat::new(secret, cleartext),
+                None => Err(PakeError::SerializationError.into()),
+            },
+            None => Err(PakeError::SerializationError.into()),
+        }?;
```
huitseeker, line 186:
```suggestion
        let sc = secret_credentials
            .iter()
            .map(|x| u8_to_credential_type(*x).ok_or(PakeError::SerializationError))
            .collect::<Result<Vec<CredentialType>, _>>()?;
        let cc = cleartext_credentials
            .iter()
            .map(|x| u8_to_credential_type(*x).ok_or(PakeError::SerializationError))
            .collect::<Result<Vec<CredentialType>, _>>()?;
        let ecf = EnvelopeCredentialsFormat::new(sc, cc)?;
```

src/envelope.rs
```diff
@@ -39,14 +40,59 @@ const NONCE_LEN: usize = 32;
 pub(crate) struct Envelope<D: Hash> {
     nonce: Vec<u8>,
     ciphertext: Vec<u8>,
+    auth_data: Vec<u8>,
     hmac: GenericArray<u8, <D as Digest>::OutputSize>,
 }
 
+pub(crate) struct OpenedEnvelopeECF {
+    pub(crate) credentials_map: HashMap<CredentialType, Vec<u8>>,
```
huitseeker, line 48: I keep thinking this `HashMap` is an overkill that actually subsumes a `BitMap` indexation into a `Vec`. There are some really tiny bitmap crates out there, here's my favorite:
https://github.com/bodil/bitmaps

src/opaque.rs
```diff
@@ -28,39 +34,25 @@ use zeroize::Zeroize;
...
-        let arr = GenericArray::from_slice(checked_slice);
-        let alpha = Grp::from_element_slice(arr)?;
-        Ok(Self { alpha })
-    }
-}
-
-impl<Grp: Group> RegisterFirstMessage<Grp> {
-    /// byte representation for the registration request
-    fn to_bytes(&self) -> GenericArray<u8, Grp::ElemLen> {
-        self.alpha.to_arr()
```
huitseeker, line 54: Why lose this one?

src/opaque.rs
```diff
@@ -104,37 +112,39 @@ where
...
 
 impl<Grp> RegisterSecondMessage<Grp>
 where
     Grp: Group,
 {
-    /// byte representation for the registration response message
-    fn to_bytes(&self) -> Vec<u8> {
-        self.beta.to_arr().to_vec()
-    }
-
```
huitseeker, line 128: lost method

src/opaque.rs
```diff
@@ -104,37 +112,39 @@ where
...
+            .map(|&x| x as u8 + 1)
+            .collect();
+        let cleartext_credentials: Vec<u8> = self
+            .ecf
+            .cleartext_credentials
+            .iter()
+            .map(|&x| x as u8 + 1)
+            .collect();
+        let ecf_serialized = [
+            serialize(secret_credentials, 1),
```
huitseeker, line 143: Two length-prefixed lists of credential types, followed by the payload are worrisome, in that they don't follow the TLS convention of `[(length-prefix, payload), ... (length-prefix, payload), ]`. Now I'm ok with that in so far as we're introducing a new and fancy type, but as long as we're going there ... might as well go with a bitmap.

huitseeker (commented):
1. Given the added complexity, are we sure everybody is going to want to follow the serialization standard? If no, wouldn't we want to keep clean variants for `to_bytes()` and `from_bytes()`?
2. Concept Ack, but this is going to need more exhaustive testing given the complexity.

src/envelope.rs
kevinlewi: Perhaps we can leave this to a future change. Created #58 to track this

src/opaque.rs
kevinlewi: Can you clarify? I am not sure I understand, I believe this is following the spec as dictated here: https://github.com/cfrg/draft-irtf-cfrg-opaque/blame/master/draft-irtf-cfrg-opaque.md#L569-L570

What is the deviation?

src/opaque.rs
kevinlewi: I think we should simply disallow a way to get bytes without calling serialize()

src/opaque.rs
kevinlewi: Same as above (I think we should simply disallow a way to get bytes without calling serialize())

kevinlewi:
Incorporated comments, and made serialization::serialize take a reference

kevinlewi:
Restored the to_bytes functions

src/opaque.rs
huitseeker: I was confused, what you're dealing with here is not a list, but a single credential formats declaration. All is right with the universe.

huitseeker (approved):
Awesome! thanks a lot, that was a lot of work!

PR #64, opened 2020-10-21
Make Hash module public

### LINKED ISSUE #63: Make "hash" module public
Hi,

Congrats for this great library!
Is it possible to make the `hash` module public? 
I already implemented my slow hash with Argon2 and do not wan't to use a forked version of this library.

Kind Regards,
Tomé

huitseeker:
Fixes #63


kevinlewi:
I think we need top-level documentation for the hash module now that it is public

kevinlewi (approved):
Thanks! cc @tomvardasca

PR #67, opened 2020-10-23
Adding simple login example and fixing some docs

kevinlewi:
This adds a simple login example via command-line interface and also fixes some extraneous doc lines in lib.rs

### DIFF
--- a/examples/simple_login.rs
+++ b/examples/simple_login.rs
@@ -1,0 +1,230 @@
+// Copyright (c) Facebook, Inc. and its affiliates.
+//
+// This source code is licensed under the MIT license found in the
+// LICENSE file in the root directory of this source tree.
+
+//! Demonstrates a simple client-server password-based login protocol
+//! using OPAQUE, over a command-line interface
+//!
+//! The client-server interactions are executed in a three-step protocol
+//! within the account_registration (for password registration) and
+//! account_login (for password login) functions. These steps
+//! must be performed in the specific sequence outlined in each of these
+//! functions.
+//!
+//! The CipherSuite trait allows the application to configure the
+//! primitives used by OPAQUE, but must be kept consistent across the steps
+//! of the protocol.
+//!
+//! In a more realistic client-server interaction, the client must send
+//! messages over "the wire" to the server. These bytes are serialized
+//! and explicitly annotated in the below functions.
+
+use rand_core::OsRng;
+use rustyline::error::ReadlineError;
+use rustyline::Editor;
+use std::collections::HashMap;
+use std::convert::TryFrom;
+use std::process::exit;
+
+use opaque_ke::{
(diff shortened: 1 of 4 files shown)


src/lib.rs
```diff
@@ -53,7 +53,7 @@
 //! # }
```
huitseeker, line 53: While I'm OK with `.unwrap()` for the hidden (`# `-prefixed) doc comments, we might want to *not* encourage unmanaged panics in visible demo lines?

examples/simple_login.rs
```diff
@@ -0,0 +1,203 @@
...
+    ciphersuite::CipherSuite,
+    keypair::KeyPair,
+    opaque::{
+        ClientLogin, ClientRegistration, LoginFirstMessage, LoginSecondMessage, LoginThirdMessage,
+        RegisterFirstMessage, RegisterSecondMessage, RegisterThirdMessage, ServerLogin,
+        ServerRegistration,
+    },
+};
+
+// The ciphersuite trait allows to specify the underlying primitives
```
huitseeker, line 21: How about an initial comment to describe the point of the example above here?

examples/simple_login.rs
```diff
@@ -0,0 +1,203 @@
...
+    // Server sends l2_bytes to client
+
+    let result = client_state.finish(
+        LoginSecondMessage::try_from(&l2_bytes[..]).unwrap(),
+        &server_kp.public(),
+        &mut client_rng,
+    );
+
+    if result.is_err() {
+        // Note that at this point, the client knows whether or not the login
```
huitseeker, line 102: Since the only sequencing bit in this example is in the `main()`, would it be possible to instead place the description of this logic there?

huitseeker (approved):
This looks great overall! I left a few nits to improve context, but they shouldn't be merge-breaking.

src/lib.rs
kevinlewi: OK, fair enough!

examples/simple_login.rs
kevinlewi: Isn't that already taken care of by the top-level comment at the top of this file?

examples/simple_login.rs
kevinlewi: Good point!

examples/simple_login.rs
huitseeker: I was expecting one more line on the degree of variability you can and should expect in your implementation. Here are things that could improve what we have now:
- where the client/server boundary is w.r.t the CLI, and in particular, where is the metaphorical "wire", and does the "get_username_and_password" function run on one side or the other?
- you can change all elements in the CipherSuite, but not in-between sub-flows
- each sub-flow has 3-steps, that must be performed in a specific sequence

PR #69, opened 2020-10-28
Add key scheduling / info computation changes for TripleDH

kevinlewi:
Our AKE implementation for TripleDH was slightly out of sync with the spec, where the spec introduced some functions for deriving the keys differently. This diff corrects those discrepancies.


src/key_exchange/tripledh.rs
```diff
@@ -387,44 +418,70 @@ fn derive_3dh_keys<KeyFormat: KeyPair, D: Hash>(
...
+        &serialize(&server_s_pk.to_arr(), 2),
     ]
     .concat();
 
-    const OUTPUT_SIZE: usize = 32;
-    let mut okm = [0u8; 3 * OUTPUT_SIZE];
-    let h = Hkdf::<D>::new(None, &ikm);
-    h.expand(&info, &mut okm)
-        .map_err(|_| InternalPakeError::HkdfError)?;
+    // TODO: could be made more efficient with re-using ikm
```
huitseeker, line 428: Is it just a matter of `let initial_hash = {let mut h = D::new();
   ... foo.update(bar); ... h};` with a `derive_secrets(initial_secret: D, ...)` that
   uses a `mut initial_secret.clone()`?

src/key_exchange/tripledh.rs
```diff
@@ -387,44 +418,70 @@ fn derive_3dh_keys<KeyFormat: KeyPair, D: Hash>(
...
-}
+    let mut hkdf_label: Vec<u8> = Vec::new();
+    hkdf_label.extend_from_slice(&length.to_be_bytes()[6..]);
 
-impl<HashLen: ArrayLength<u8>> TryFrom<&[u8]> for KE3Message<HashLen> {
-    type Error = InternalPakeError;
+    let mut opaque_label: Vec<u8> = Vec::new();
+    opaque_label.extend_from_slice(&STR_OPAQUE);
+    opaque_label.extend_from_slice(&label);
+    hkdf_label.extend_from_slice(&serialize(&opaque_label, 1));
```
huitseeker, line 465: Do we want to hash in the length-prefixed serialized format, or simply the
bits themselves?
Edit: Answering my own question, [we want the serialization](https://github.com/cfrg/draft-irtf-cfrg-opaque/blob/master/draft-irtf-cfrg-opaque.md#3dh-key-derivation-derive-3dh).

huitseeker (commented):
LGTM, modulo the easy-to-fix TODO

src/key_exchange/tripledh.rs
kevinlewi: Thanks, yes you are right, I addressed this TODO in an update to this PR.

huitseeker (approved):
Thanks \o/

Issue #33, opened 2020-07-27
Adopt RFC names


huitseeker:
We rely on number to name the messages of our protocol:
```
huitseeker@bucephalus?huitseeker/tmp/opaque-ke(master)? rg '(fn|struct) \w+[0-9]\w*\b'                                                                                                                                                                                                     [13:49:17]
benches/oprf.rs
17:fn oprf1(c: &mut Criterion) {
31:fn oprf1_edwards(c: &mut Criterion) {
45:fn oprf2(c: &mut Criterion) {
66:fn oprf2_edwards(c: &mut Criterion) {
87:fn oprf3(c: &mut Criterion) {
110:fn oprf3_edwards(c: &mut Criterion) {

src/oprf.rs
20:pub(crate) fn generate_oprf1<R: RngCore + CryptoRng, G: GroupWithMapToCurve>(
36:pub(crate) fn generate_oprf2<G: Group>(
45:pub(crate) fn generate_oprf3<G: Group>(
59:pub fn generate_oprf1_shim<R: RngCore + CryptoRng, G: GroupWithMapToCurve>(
69:pub fn generate_oprf2_shim<G: Group>(
78:pub fn generate_oprf3_shim<G: Group>(

src/tests/mock_rng.rs
39:    fn next_u32(&mut self) -> u32 {
44:    fn next_u64(&mut self) -> u64 {

src/key_exchange/traits.rs
21:    fn generate_ke1<R: RngCore + CryptoRng, KeyFormat: KeyPair<Repr = Key>>(
26:    fn generate_ke2<R: RngCore + CryptoRng, KeyFormat: KeyPair<Repr = Key>>(
35:    fn generate_ke3<KeyFormat: KeyPair<Repr = Key>>(
48:    fn ke1_state_size() -> usize;
50:    fn ke2_message_size() -> usize;

src/keypair.rs
208:pub struct X25519KeyPair {
282:        fn test_x25519_check(kp in X25519KeyPair::uniform_keypair_strategy()) {
288:        fn test_x25519_pub_from_priv(kp in X25519KeyPair::uniform_keypair_strategy()) {
296:        fn test_x25519_dh(kp1 in X25519KeyPair::uniform_keypair_strategy(),

src/tests/opaque_ke_test.rs
24:struct X255193dhNoSlowHash;
342:fn test_r1() -> Result<(), PakeError> {
360:fn test_r2() -> Result<(), PakeError> {
[...]

PR #71, opened 2020-11-02
Some changes in order to conform to VOPRF spec

kevinlewi:
- Removing pepper parameter
- Adding Token struct used in VOPRF RFC
- Got rid of blinding_factor_raw parameter, added id_u and id_s to parameters for test vectors as well


kevinlewi:
Added 3 TODOs, each tracked with an issue: #72, #73, #74

huitseeker:
This is a good 1st step, thanks!
I note this is very close to #33 and to make sure we don't duplicate any work, our first priority should be to collapse the master & TLS branches.

huitseeker (approved):

PR #77, opened 2020-11-03
Start integrating the derivable SizedBytes

huitseeker:
- just derive it on Key and KeyPair for now.

 Eliminates a complex derivation.


kevinlewi (approved):
LGTM

PR #78, opened 2020-11-04
Uses conditional compilation to make sure postprocessing is only avai…

huitseeker:
…lable in tests

See
https://github.com/rust-lang/rust/pull/64010/


kevinlewi (approved):
Neat trick!

Issue #72, opened 2020-11-02
Incorporate expand_message_xmd from hash-to-curve RFC


kevinlewi:
The map_to_curve function currently does a SHA512 evaluation. We need to instead implement and use the expand_message_xmd function from https://www.ietf.org/archive/id/draft-irtf-cfrg-hash-to-curve-10.txt.

huitseeker:
Yep!
Domain separation strikes again.

PR #80, opened 2020-11-06
Adding expand_message_xmd and RFC compliance with VOPRF and H2C

kevinlewi:
Addresses #72, #73, and #74 

Note that x25519 is not in the VOPRF spec, and so we actually don't have anything to match against... but this will handle ristretto255 at least.


src/map_to_curve.rs
```diff
@@ -6,31 +6,234 @@
...
+        Ok(<Self as Group>::hash_to_curve(&hashed_input))
+    }
+}
+
+// Computes ceil(x / y)
+fn div_ceil(x: usize, y: usize) -> usize {
+    let additive = match x % y != 0 {
+        true => 1,
+        false => 0,
+    };
```
huitseeker, line 62: Would `if x % y == 0 { 0 } else { 1 }` be simpler? (or if we're really going for it, `(x%y != 0) as usize`)

src/map_to_curve.rs
```diff
@@ -6,31 +6,234 @@
...
+#[cfg(test)]
+mod tests {
+
+    struct Params {
+        msg: &'static str,
+        len_in_bytes: usize,
+        uniform_bytes: &'static str,
+    }
+
+    #[test]
```
huitseeker, line 126: Super!

src/serialization/mod.rs
```diff
@@ -80,23 +80,43 @@ impl<T: CipherSuite> From<&LoginThirdMessage<T>> for ProtocolMessageType {
...
 }
 
-pub(crate) fn serialize(input: &[u8], max_bytes: usize) -> Vec<u8> {
-    let mut output: Vec<u8> = Vec::new();
-    output.extend_from_slice(&input.len().to_be_bytes()[8 - max_bytes..]);
-    output.extend_from_slice(&input[..]);
+// Corresponds to the I2OSP() function from RFC8017
+pub(crate) fn i2osp(input: usize, length: usize) -> Vec<u8> {
+    if length <= 8 {
+        return (&input.to_be_bytes()[8 - length..]).to_vec();
```
huitseeker, line 86: If usize is 32 bytes on a particular platform, and I want to write it on len=4
bytes, it seems I'd be writing .. nothing? `std::mem::size_of::<usize>()` may help here.

src/serialization/mod.rs
```diff
@@ -80,23 +80,43 @@ impl<T: CipherSuite> From<&LoginThirdMessage<T>> for ProtocolMessageType {
...
 
+// Corresponds to the OS2IP() function from RFC8017
+pub(crate) fn os2ip(input: &[u8]) -> Result<usize, PakeError> {
+    if input.len() > 8 {
+        return Err(PakeError::SerializationError);
+    }
+
+    let mut output_array = [0u8; 8];
+    for i in 0..input.len() {
+        output_array[8 - input.len() + i] = input[i];
```
huitseeker, line 102: Would it be simpler to use `copy_from_slice`, considering it takes a `&mut self`?

src/serialization/mod.rs
```diff
@@ -80,23 +80,43 @@ impl<T: CipherSuite> From<&LoginThirdMessage<T>> for ProtocolMessageType {
...
+    }
+    Ok(usize::from_be_bytes(output_array))
+}
```
huitseeker, line 105: Could we unit-test i2osp, is2ip, preferably with proptest and potentially with a cross-test on u32? I'm happy adding 32-bits compilation / testing to CI

huitseeker:
@kevinlewi See also the run results on #81 which show we already have issues in this area
https://github.com/novifinancial/opaque-ke/pull/81/checks?check_run_id=1374926279

src/serialization/mod.rs
huitseeker: CI is in #81

src/serialization/mod.rs
kevinlewi: Changed "8" to std::mem::size_of::<usize>() everywhere

src/serialization/mod.rs
kevinlewi: Added a proptest:

```
#[test]
fn test_i2osp_os2ip(bytes in vec(any::<u8>(), 0..std::mem::size_of::<usize>())) {
    assert_eq!(i2osp(os2ip(&bytes)?, bytes.len()), bytes);
}
```

huitseeker (approved):
Awesome!

PR #85, opened 2020-11-13
Updating AKE message format

kevinlewi:
Includes the changes which add the info and Einfo parameters to the serialization of the AKE messages.

Things that still need to be addressed:
- idU and idS should be taken from the envelope, and then fall back to pkU and pkS if they are not available in the envelope
- the generate_ke2/3 functions should return identities extracted from the key exchange protocol, plus the info and einfo fields
- einfo needs to be encrypted under Ke2/Ke3


huitseeker (approved):
nit: let's not percolate all errors to the level of `PakeError`

Issue #84, opened 2020-11-12
simple_login example panics when registering user


0x6273:
```thread 'main' panicked at 'called `Result::unwrap()` on an `Err` value: VerificationError(CryptoError(SizeError { name: "first_message_bytes", len: 32, actual_len: 38 }))', examples/simple_login.rs:66:55```

I was able to get the example to work with the following changes:
```diff
diff --git a/examples/simple_login.rs b/examples/simple_login.rs
index 50f0f0c..4eee1af 100644
--- a/examples/simple_login.rs
+++ b/examples/simple_login.rs
@@ -63,7 +63,7 @@ fn account_registration(
 
     let mut server_rng = OsRng;
     let (r2, server_state) = ServerRegistration::<Default>::start(
-        RegisterFirstMessage::try_from(&r1_bytes[..]).unwrap(),
+        RegisterFirstMessage::deserialize(&r1_bytes[..]).unwrap(),
         &mut server_rng,
     )
     .unwrap();
@@ -73,7 +73,7 @@ fn account_registration(
 
     let (r3, _) = client_state
         .finish(
-            RegisterSecondMessage::try_from(&r2_bytes[..]).unwrap(),
+            RegisterSecondMessage::deserialize(&r2_bytes[..]).unwrap(),
             server_kp.public(),
             &mut client_rng,
         )
@@ -83,7 +83,7 @@ fn account_registration(
     // Client sends r3_bytes to server
 
     let password_file = server_state
-        .finish(RegisterThirdMessage::try_from(&r3_bytes[..]).unwrap())
+        .finish(RegisterThirdMessage::deserialize(&r3_bytes[..]).unwrap())
         .unwrap();
     password_file.to_bytes()
 }
@@ -106,7 +106,7 @@ fn account_login(
     let (l2, server_state) = ServerLogin::start(
         password_file,
         &server_kp.private(),
-        LoginFirstMessage::try_from(&l1_bytes[..]).unwrap(),
+        LoginFirstMessage::deserialize(&l1_bytes[..]).unwrap(),
         &mut server_rng,
     )
     .unwrap();
@@ -115,7 +115,7 @@ fn account_login(
     // Server sends l2_bytes to client
 
     let result = client_state.finish(
-        LoginSecondMessage::try_from(&l2_bytes[..]).unwrap(),
+        LoginSecondMessage::deserialize(&l2_bytes[..]).unwrap(),
         &server_kp.public(),
         &mut client_rng,
     );
[...]

PR #86, opened 2020-11-13
Fixing simple_login example

kevinlewi:
See #84


kevinlewi:
@huitseeker is there any way we can add automated checks via github actions in order to verify that our examples work? simple_login may be non-trivial since it accepts input from stdin.

huitseeker (approved):
@kevinlewi Addressed in #87

PR #91, opened 2020-11-19
Adding test vector generation example

kevinlewi:
Moving test vector generation out of tests and into a separate file in examples/

Also adding a feature "test-vector" in order to make this work


huitseeker (changes requested):
- a number of files change mode and become executable for all (100644 -> 100755), this does not seem necessary,
- rather than an `example`, which does not seem to have its natural place in the repo, we could run `test_vector_generation` [as a build script](https://blog.cyplo.dev/posts/2018/12/generate-rust-tests-from-data/) of a(n integration) test crate,
- the test-vector feature exposes risky getters, and the gates using that feature are not and-gated with `cfg(test)` either 
- is there a reason we want the test vector generation outside the tests module?

PR #92, opened 2020-11-29
Adding info and einfo fields to test vectors

kevinlewi:
Adds info1, info2, einfo2, info3, einfo3 to test vector generation


huitseeker (approved):
Super!

I'd be ok with either removing the wrap clippy signals, or adding a local `#[allow(clippy:unnecessary_wraps)]` on the `oprf::evaluate`.

PR #93, opened 2020-11-30
Removing protocol message bytes

kevinlewi:
Following https://github.com/cfrg/draft-irtf-cfrg-opaque/commit/a274e8e6f7fbefeaaae762510bb63c7869e73c63


huitseeker (approved):
Super!

PR #95, opened 2020-12-04
Adding VOPRF draft's test vectors

kevinlewi:
This ensures that our oprf implementation is compatible with [the VOPRF draft](https://github.com/cfrg/draft-irtf-cfrg-voprf/blob/master/draft-irtf-cfrg-voprf.md), currently version 06.

Test vectors are copied for the "base mode" of VOPRF for Ristretto255-SHA512.


huitseeker (approved):
:+1:

Issue #97, opened 2020-12-08
Add description of optional parameters to documentation


kevinlewi:
The optional parameters need to be better-documented, but this should wait until they are settled from the RFC side first.

Adding as a note to myself for future reference:

```
//! ## Optional Fields
//!
//! * the export key `export_key_login` which will match `export_key_registration` if the same password was entered,
//! * the server's public key `server_s_pk`,
//! * an optional server identifier `id_s`,
//! * a `plain_info` field optionally set as the `info` field in an optional parameter to `credential_response`, and
//! * a `confidential_info` field optionally set as the `confidential_info` field in an optional parameter to `credential_response`.
```

PR #98, opened 2020-12-08
Cleaning up docs + README

kevinlewi:
- README changes, update links to most recent draft cfrg-01
- rename everything from r1/r2/r3 to RFC names
- remove all mentions of pepper, client supplying server_s_pk
- make rng the first argument of every high-level API function
- registration outputs should return tuple types

Omitted all optional parameters from the main documentation for now, but will add a separate section (see issue #97) for describing how to use them once the internet draft settles on these issues


huitseeker (approved):
Super nice cleanup, thanks! \o/
- Should we close #33 since we're pretty much making the expected changes here? 
- given how long the block in `src/lib` is getting, do we want to add an issue to track [upstream](https://github.com/rust-lang/rust/issues/44732) before splitting this into another file?

Issue #104, opened 2020-12-27
Allow providing (idU, idS) at ClientLoginFinish phase


PaulGrandperrin:
For the same reasons evoked in #100 and in the continuation of #102 , it would be nice to allow providing (idU, idS) at ClientLoginFinish phase.

kevinlewi:
Thanks again for pointing this out -- I forgot to address it in the previous PR. I will roll this out along with some [other changes coming to the envelope construction](https://github.com/cfrg/draft-irtf-cfrg-opaque/pull/102) (which will alter the API slightly)

Issue #58, opened 2020-10-08
Change HashMap usage to BitMap indexing for credentials format


kevinlewi:
YMMV, but I keep thinking this `HashMap` is an overkill that actually could be replaced by a `BitMap` indexation into a `Vec`. There are some really tiny bitmap crates out there, here's my favorite:
https://github.com/bodil/bitmaps

_Originally posted by @huitseeker in https://github.com/novifinancial/opaque-ke/pull/56#discussion_r499942393_

PR #108, opened 2020-12-30
Updating envelope structure to support two fixed modes

kevinlewi:
To keep in sync with the changes in https://github.com/cfrg/draft-irtf-cfrg-opaque/pull/102.

The envelope will now support only two modes: a base mode, and a custom identifier mode. This means that we can get rid of the envelope credentials format that we were previously using, which also makes #58 obsolete.

This also addresses #104.


kevinlewi:
@huitseeker for review

src/envelope.rs
```diff
@@ -305,13 +277,30 @@ impl<D: Hash> Envelope<D> {
...
             plaintext,
             export_key: *GenericArray::from_slice(&export_key),
         })
     }
 }
 
+// Helper functions
+
+fn construct_aad(server_s_pk: &[u8], optional_ids: &Option<(Vec<u8>, Vec<u8>)>) -> Vec<u8> {
+    let serialized_ids = match optional_ids {
```
huitseeker, line 290: I believe this is `optional_ids.iter().flat_map(|(l, r)| [serialize(l, 2), serialize(r, 2)].concat()).collect()` [(playground)](https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=98ba4f0ce5c122d1d3034a8ff7488839)

src/messages.rs
```diff
@@ -319,15 +274,16 @@ impl<CS: CipherSuite> CredentialResponse<CS> {
...
         credential_response.extend_from_slice(&self.ke2_message.to_bytes());
         credential_response
     }
 
     /// Deserialization from bytes
     pub fn deserialize(input: &[u8]) -> Result<Self, ProtocolError> {
-        let (beta_bytes, envelope_and_ke2m_bytes) = tokenize(&input, 2)?;
-        let concatenated = [&beta_bytes[..], &envelope_and_ke2m_bytes[..]].concat();
+        let (beta_bytes, remainder) = tokenize(&input, 2)?;
+        let concatenated = [&beta_bytes[..], &remainder[..]].concat();
```
huitseeker, line 286: nit: I get we have more cases to cover, but can we still find a more descriptive name than "remainder"?

huitseeker (approved):
Super! I left just one reformulation

src/envelope.rs
```diff
@@ -305,13 +277,29 @@ impl<D: Hash> Envelope<D> {
...
 }
 
+// Helper functions
+
+fn construct_aad(server_s_pk: &[u8], optional_ids: &Option<(Vec<u8>, Vec<u8>)>) -> Vec<u8> {
+    optional_ids
+        .iter()
+        .flat_map(|(l, r)| [serialize(server_s_pk, 2), serialize(l, 2), serialize(r, 2)].concat())
+        .collect()
+}
```
huitseeker, line 294: I'm afraid this will need an adaptation, since we would still like to get `[serialize(server_s_pk, 2)]` when `optional_ids` is `None`. 

What I meant to suggest was :
```
let server_ids = optional_ids
        .iter()
        .flat_map(|(l, r)| [serialize(l, 2), serialize(r, 2)].concat())
        .collect();
[serialize(server_s_pk, 2), server_ids].concat()
 ```

kevinlewi: Oops, thanks for the catch! #112

Issue #107, opened 2020-12-29
Make public the documentation of structs being part of the API, like `ClientLoginFinishResult`


PaulGrandperrin:
In #105 , I have been pointed to the `export_key` field of `ClientLoginFinishResult` and `ClientRegistrationFinishResult` but none of those structs are available in the documentation (on doc.rs(0.2) and from master).

Those structs and their fields are already `pub` (and therefore available to the library users) but because they are not `pub use` in `lib.rs`, they are not exposed in the documentation.

I think a choice should be made between:
- making those structs and fields officially part of the API by making them visible in the doc
- removing them completely from the API surface and instead providing public and documented accessors

The same question can be made about `ClientLoginStartResult` and `ClientRegistrationStartResult` and probably others.

PR #110, opened 2021-01-03
Adding documentation for advanced usage + optional parameters

kevinlewi:
To be landed after #108 (since it is a dependency). This adds documentation for the result structs and all of the "advanced features", namely:
- usage of the shared secret
- export key
- checking server_s_pk
- custom identifiers
- plain_info and confidential_info in key exchange

Addresses #107 and #97

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -65,7 +65,7 @@
 //!
 //! ### Client Registration Start
 //! In the first step of registration, the client chooses as input a registration password. The client runs [ClientRegistration::start]
-//! to produce an output consisting of a [RegistrationRequest] to be sent to the server, and
+//! to produce a [ClientRegistrationStartResult], which consists of a [RegistrationRequest] to be sent to the server and
 //! a [ClientRegistration] which must be persisted on the client for the final step of client registration.
 //! ```
 //! # use opaque_ke::{
@@ -96,8 +96,8 @@
 //! ### Server Registration Start
 //! In the second step of registration, the server takes as input the instance of [RegistrationRequest] from the client, and
 //! the server's public key `server_kp.public()`.
-//! The server runs [ServerRegistration::start] to produce an output consisting of
-//! a [RegistrationResponse] to be returned to the client, and
+//! The server runs [ServerRegistration::start] to produce an a [ServerRegistrationStartResult], which consists of
+//! a [RegistrationResponse] to be returned to the client and
 //! a [ServerRegistration] which must be persisted on the server for the final step of server registration.
 //! ```
 //! # use opaque_ke::{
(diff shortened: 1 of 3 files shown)


src/lib.rs
```diff
@@ -404,16 +400,341 @@
...
+//! # Advanced Usage
+//!
+//! This implementation offers support for several optional features of OPAQUE, described below. They are not critical to the
+//! execution of the main protocol, but can provide additional security benefits which can be suitable for various applications that rely on
+//! OPAQUE for authentication.
+//!
+//! ## Shared Secret
+//!
+//! Upon a successful completion of the OPAQUE protocol (the client runs login with the same password used during registration),
+//! the client and server have access to a shared secret, which is a pseudorandomly distributed
```
huitseeker, line 425: nit: line wrap

src/lib.rs
```diff
@@ -404,16 +400,341 @@
...
+//! produce different shared secrets, distributed as uniformly random strings. Thus, the shared secret can be used as a session secret
+//! for a secure channel between the client and server.
+//!
+//! The shared secret can be accessed from the `shared_secret` field of [ClientLoginFinishResult] and [ServerLoginFinishResult]. See
+//! the combination of [Client Login Finish](#client-login-finish) and [Server Login Finish](#server-login-finish) for example usage.
+//!
+//! ## Checking Server Consistency
+//!
+//! A [ClientLoginFinishResult] contains the `server_s_pk` field, which represents the static public key of the server that is established
+//! during the setup phase. This can be used by the client to verify the authenticity of the server it engages with during the login phase. In particular,
```
huitseeker, line 436: nit: something like : "contains the `server_s_pk`, which is extracted from ..." or "As this value is extracted from the envelope generated at registration ..."

src/lib.rs
```diff
@@ -404,16 +400,341 @@
...
+//! public key beforehand.
+//!
+//!
+//! ## Export Key
+//!
+//! The export key is a pseudorandomly distributed 32-byte string output by both the
+//! [Client Registration Finish](#client-registration-finish) and [Client Login Finish](#client-login-finish) steps.
+//! The same export key string will be output by both functions only if the exact same password is passed to [ClientRegistration::start] and [ClientLogin::start].
+//!
+//! The export key retains as much secrecy as the password itself. Hence, only the parties which
```
huitseeker, line 508: nit: An allusion to (optional) password hashing being in the pipeline between key and password would give more context as to the feasibility of brute-force attacks. Even better if we're able to refer to another section of our doc for details on export key derivation.

huitseeker (approved):

src/lib.rs
kevinlewi: Technically with the new envelope structure, it is actually not extracted from the envelope. It instead must be supplied in order to verify the mac on the envelope, but the value itself isn't present in the envelope.

huitseeker: :eyes: Thanks!

PR #116, opened 2021-01-15
Adding a from_private_key_slice on KeyPair

### LINKED ISSUE #109: Provide helper to load a CryptoSuite Keypair from &[u8]
At the moment, it's quite easy to gen a new random keypair:
```rust
let kp = <common::crypto::OpaqueConf as CipherSuite>::generate_random_keypair(&mut rng).unwrap();
```

It's also quite easy to serialize it, i think because `keypair::Key` implements `Deref<Target=Vec<u8>>`:
```rust
let mut f = async_std::fs::File::create(common::consts::OPAQUE_PRIVATE_KEY_PATH).await?;
f.write_all(kp.private());
```

but it is very complicated to load it from a slice:
```rust
let mut f = File::open(common::consts::OPAQUE_PRIVATE_KEY_PATH).await?;
let mut sk = Vec::new();
f.read_to_end(&mut sk).await?;

let sk = generic_array::GenericArray::from_slice(&sk);
let sk = <<<common::crypto::OpaqueConf as CipherSuite>::KeyFormat as KeyPair>::Repr as SizedBytes>::from_arr(sk).unwrap();
        
let pk = <<common::crypto::OpaqueConf as CipherSuite>::KeyFormat as KeyPair>::public_from_private(&sk);
let kp = <<common::crypto::OpaqueConf as CipherSuite>::KeyFormat as KeyPair>::new(pk, sk)?;
```
and it also requires adding `generic-bytes` and `generic-array` as dependencies.

There might be an easier way going through the `Key` type which implements `TryFrom` but I haven't found it.

kevinlewi:
Closes #109 

@PaulGrandperrin, you should now be able to do something like:

```
let kp = X25519KeyPair::from_private_key_slice(&private_key_bytes)?;
```

### DIFF
--- a/src/keypair.rs
+++ b/src/keypair.rs
@@ -31,7 +31,7 @@ pub(crate) trait SizedBytesExt: SizedBytes {
 impl<T> SizedBytesExt for T where T: SizedBytes {}
 
 /// A Keypair trait with public-private verification
-pub trait KeyPair: Sized {
+pub trait KeyPair: Sized + Clone {
     /// The single key representation must have a specific byte size itself
     type Repr: SizedBytes + Clone;
 
@@ -60,6 +60,13 @@ pub trait KeyPair: Sized {
 
     /// Computes the diffie hellman function on a public key and private key
     fn diffie_hellman(pk: Self::Repr, sk: Self::Repr) -> Vec<u8>;
+
+    /// Obtains a KeyPair from a slice representing the private key
+    fn from_private_key_slice(input: &[u8]) -> Result<Self, InternalPakeError> {
+        let sk = Self::Repr::from_arr(GenericArray::from_slice(&input))?;
+        let pk = Self::public_from_private(&sk);
+        Self::new(pk, sk)
+    }
 }
 
 #[cfg(test)]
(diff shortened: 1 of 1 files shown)


PaulGrandperrin:
Yes, that works great!

Also, maybe it is not the correct place to talk about it but it is really minor and related: the type `Keypair` has the following usability issue:
- On the server-side, one wants to keep the `Keypair` available in a shared state, available to all request handlers.
- For example, using `tide`, it would mean storing the `Keypair` in the state: https://docs.rs/tide/0.15.0/tide/fn.with_state.html
- Unfortunately, this requires that the state be `Clone` but `Keypair` is not:
```
error[E0277]: the trait bound `X25519KeyPair: Clone` is not satisfied
 --> server/src/state.rs:9:5
  |
9 |     pub opaque_kp: <common::crypto::OpaqueConf as CipherSuite>::KeyFormat,
  |     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ the trait `Clone` is not implemented for `X25519KeyPair`
  |
  = note: required by `clone`
  = note: this error originates in a derive macro (in Nightly builds, run with -Z macro-backtrace for more info)
```
- adding `derive(Clone)` to the needed structures in `opaque-ke` solves the issue

kevinlewi:
Cool, sounds reasonable -- adding a `derive(Clone)` to `KeyPair`

PR #127, opened 2021-01-31
Addressing various TODOs and minor cleanups

### LINKED ISSUE #124: Consider renaming OpenedEnvelope struct
To be something more clear, since it also provides the export key (which is technically unrelated to the envelope)

Suggestion from @haochenuw

### LINKED ISSUE #126: Digital locker example should encrypt payload in last step of login
See title. Should use the shared secret to encrypt

kevinlewi:
- Removing use of NONCE_SIZE and using NonceLen type instead
- Fixing digital locker example
- Modularizing the serialization of various messages when used in AKE section

Closes #124 and #126

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -390,12 +390,6 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "bb1f6b1ce1c140482ea30ddd3335fc0024ac7ee112895426e0a629a6c20adfe3"
 
 [[package]]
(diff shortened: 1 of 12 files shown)


huitseeker (approved):
Thanks! This looks great, thanks for the rename of `NONCE_LEN` in particular.

nit: there's an `.unwrap()` in the initialization of `client_login_finish_result` in `test_credential_finalization` that could be a `?`

PR #129, opened 2021-02-05
Reuse hashers by `hasher.finalize_reset()`

kchalkias:
This is preferred (especially in loops) as it's sometimes faster than reconstructing the hasher object.
See [function's docs](https://docs.rs/digest/0.9.0/digest/trait.Digest.html#tymethod.finalize_reset)


kevinlewi (approved):
Thank you!

PR #130, opened 2021-02-05
Fixing envelope mode inconsistency and matching against latest test vectors

kevinlewi:
Matches test vectors against https://github.com/cfrg/draft-irtf-cfrg-opaque/pull/138


huitseeker (approved):
I see. Nice that we now have test vectors to compare to!

PR #103, opened 2020-12-22
Update rand to 0.8 and rand_core to 0.6

PaulGrandperrin:
This makes a better user experience when targeting wasm.
Closes #101

PR #131, opened 2021-02-05
Updating rand_core to 0.6 (without curve25519-dalek library change)

### LINKED ISSUE #101: Update rand_core to 0.6
From @PaulGrandperrin in #100:

> Also as a small aside, it would be nice if you could wait for rand_core to be released in version 0.6 (rust-random/rand#1072) before releasing your 0.3 version. The new version makes it possible to use rand_core::OsRng on wasm.

We however need to also wait for curve25519-dalek + other dependencies to upgrade their `rand_core` dependency.

cc: @huitseeker

kevinlewi:
Same goal as #103, but without adjusting any curve25519-dalek library dependencies.

This is because we can just rely on calling the internals of `Scalar::random` in the one place where it is used.

Closes #101 

cc: @PaulGrandperrin

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -338,7 +338,7 @@ checksum = "c8492de420e9e60bc9a1d66e2dbb91825390b738a388606600663fc529b4b307"
 dependencies = [
  "byteorder",
  "digest",
- "rand_core",
+ "rand_core 0.5.1",
  "subtle",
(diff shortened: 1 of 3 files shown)


huitseeker (approved):
Yep, definitely the cleaner approach at this point.

PR #156, opened 2021-04-23
Zeroize keys on drop

antoniokdacunha:
Currently the main key structure is not zeroized on drop. This diff adds this best practice functionality.


kevinlewi:
This is great. Thanks @antoniokdacunha!

kevinlewi (approved):

PR #168, opened 2021-06-10
Implement serde serialization and deserialization

### LINKED ISSUE #166: Declare a trait for `serialize`/`deserialize`
Many structures in this crate can be serialized/deserialized with the same interface. However, it's all just ad-hoc type methods. It would allow writing generic code if the methods were extracted to a trait.

Even better, implementing the `serde` interface is the standard way to serialize and deserialize code.

Brolock:
Provide a standard interface for serialization and deserialization using serde.

This makes it much easier to pass messages as part of protocols.

Fixes #166


kevinlewi:
Thanks @Brolock for adding this, looks great!

PR #174, opened 2021-06-15
Implement Clone for every message type

nitnelave:
This makes it much easier to handle the messages as part of a framework (e.g. actix where in some cases move-only is not possible). Note that it was _technically_ possible so far using `serialize` and `deserialize`, but that's inefficient and ugly.

Note that just using `#[derive(Clone)]` is not possible for many of the types, since they are generic over `CS`, and deriving `Clone` would add the requirement that `CS` is `Clone`, which in turns would require `KeyExchange` to be `Clone`, which doesn't make sense (we never own an instance of it).


kevinlewi:
Awesome! Thanks.

kevinlewi (approved):

PR #175, opened 2021-06-15
Enforce public vs private keys via types

nitnelave:
With this change, it's no longer possible to pass a public key where a private one is expected, and vice-versa.

I *might* have made the mistake when using this library :)

### DIFF
--- a/src/envelope.rs
+++ b/src/envelope.rs
@@ -6,7 +6,7 @@
 use crate::{
     errors::{utils::check_slice_size_atleast, InternalPakeError, PakeError, ProtocolError},
     hash::Hash,
-    keypair::Key,
+    keypair::PublicKey,
     serialization::serialize,
 };
 use digest::Digest;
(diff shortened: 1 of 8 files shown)


kevinlewi (commented):
Looks like a great change -- thanks!

But I'd like to hold off on landing this change since it does affect the public API, and would be better reserved for v0.6.

I am currently prioritizing changes for a v0.5.1 first.

nitnelave:
If that's the case, then I might try to rework it a bit: make it also differentiated for client and server key as well, while refactoring all to be a single (generic) type

kevinlewi:
Actually, scrapping v0.5.1 plans, let's just aim for a v0.6 and bundle all of these changes together with the next branch.

Anyway, this looks ready to me, unless you wanted to make some more changes first. Let me know either way, happy to merge as-is!

nitnelave:
Hmm, after digging a bit into it, it would be very complicated to have client or server-specific keys (the diffie-hellman doesn't care and mixes them happily, for instance), and sharing the implementation with a marker complicates a bunch of `#[derive]` because of the `PhantomData` that is then required :/

I think it's as good as it's going to get like this :)

kevinlewi (approved):

PR #181, opened 2021-06-21
Argon2 implementation

### LINKED ISSUE #179: Argon2 support
Right now only support for scrypt is implemented. Is Argon2 support in scope of this project?

I have an implementation ready that is based on https://crates.io/crates/argon2, I was wondering what I should PR against, `master` or `0.6`?

daxpedda:
I removed the `slow-hash` feature and replaced it with `scrypt` and `argon`. As per spec recommendation, I chose Argon2id, but I am no cryptographer and a different variant might be desirable when used in OPAQUE.

Fixes #179.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -70,7 +70,7 @@ jobs:
     steps:
       - uses: actions/checkout@v2
       - uses: hecrj/setup-rust-action@v1
-      - run: cargo test --verbose --features scrypt,argon --no-default-features --features ${{ matrix.backend_feature }}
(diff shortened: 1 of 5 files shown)


kevinlewi:
Thanks for this! I think actually we can go ahead and just remove the existing scrypt feature / dependency / slow-hash implementation entirely and replace with Argon2id.

daxpedda:
> Thanks for this! I think actually we can go ahead and just remove the existing scrypt feature / dependency / slow-hash implementation entirely and replace with Argon2id.

Done, I renamed the feature back to `slow-hash`, should this be `argon2` instead?

kevinlewi (approved):
Looks great, thanks for the contribution!

PR #182, opened 2021-06-21
Configurable SlowHash

### LINKED ISSUE #180: SlowHash configurability
At the moment, it isn't possible to pass parameters to `SlowHash::hash()`, the only way I could figure out is to make a new type implementing `SlowHash` that just executes a different function.

Would it be in scope to pass down a instantiated `CipherSuite::SlowHash` to `SlowHash::hash()` through either an additional parameter in `ClientRegistration/ClientLogin::finish()` or adding something to `ClientRegistrationFinishParameters/ClientLoginFinishParameters`?

daxpedda:
~~I decided to re-use the existing parameters.~~
In light of the latest `master`, I added a new parameter to `finish` instead.

Fixes #180.

### DIFF
--- a/examples/digital_locker.rs
+++ b/examples/digital_locker.rs
@@ -26,6 +26,7 @@
 
 use chacha20poly1305::aead::{Aead, NewAead};
 use chacha20poly1305::{ChaCha20Poly1305, Key, Nonce};
+use opaque_ke::slow_hash::NoOpHash;
 use rustyline::error::ReadlineError;
 use rustyline::Editor;
 use std::process::exit;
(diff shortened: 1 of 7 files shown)


daxpedda:
Rebased.

daxpedda:
Fixed leftover `Default` constraint from when it was still an `Option`.

daxpedda:
Rebased.

kevinlewi:
I'm a little hesitant to pull these out of the ciphersuite configuration and into the function calls themselves. See https://github.com/novifinancial/opaque-ke/issues/180#issuecomment-865411309 for some thoughts on how we might be able to support configurable SlowHash without making any changes to the existing code -- let me know if that works for you, or if it is missing something for your use-case!

daxpedda:
I will close this, will need a new PR anyway after finishing the discussion in #180.

PR #184, opened 2021-06-22
Implement common traits

daxpedda:
This is an attempt to make sure that all exposed types implement `Clone`, `Debug`, `Deserialize` and `Serialize`. As a bonus I threw in `Eq`, `Hash` and `PartialEq` because it was no extra effort.

I skipped on implementing it on the `XxxResult` types because their public members are covered, which works great for our use-case, but adding them would be simple.

`Hash` might not be appropriate in a couple of places, I wouldn't know.


kevinlewi (approved):
Great change!

PR #185, opened 2021-06-22
Update dependencies

daxpedda:
This updates the following dependencies:
- curve25519-dalek: 3.0.0 -> 3.1.0
- hkdf: 0.10.0 -> 0.11.0
- hmac: 0.10.1 -> 0.11.0
- crypto-mac (sub-dependency): 0.10.0 -> 0.11.0

I skipped updating `displaydoc` because the [multi-line change](https://github.com/yaahc/displaydoc/blob/master/CHANGELOG.md#changed) they did is annoying.

It is [usually recommended](https://doc.rust-lang.org/cargo/guide/cargo-toml-vs-cargo-lock.html#cargotoml-vs-cargolock) that libraries don't include `Cargo.lock`, what is opaque-ke's take on this?


kevinlewi (approved):
This looks great!

Let's remove the Cargo.lock from being tracked. No strong opinion here, and it does seem like it is better to practice to not track it. Thanks for catching this!

PR #192, opened 2021-06-25
Expose public key in RegistrationResponse

daxpedda:
I hope this is correct. This would help transmitting the public key during registration, removing the need to send it separately.


daxpedda:
Replaced by #193.

PR #193, opened 2021-06-25
Expose public key in ClientRegistrationFinishResult

daxpedda:
I realized this is more in line with the style then #192.

This enables to obtain the servers public key during registration.


kevinlewi (approved):
Looks good!

PR #200, opened 2021-07-06
Group impl support

### LINKED ISSUE #198: `Group` implementation support
I have done some work towards implementing P256 support in opaque-ke. In this process I discovered some potential shortcomings in some of the interfaces:
- `PublicKey` and `PrivateKey` are currently fixed to be 32-bytes, which doesn't apply to other prime-order groups, like P256
- [p256's `Scalar` implementation](https://docs.rs/p256/0.9.0/p256/struct.Scalar.html) for example, doesn't support returning a reference to it's underlying memory, but currently `Group::scalar_as_bytes` requires this, I figured that `Group::Scalar` might just implement `Copy` to solve this issue, which might be reasonable?
- `expand_message_xmd` is currently not exposed, which makes sense, but I understand if opaque-ke expects implementors to bring their own implementation
- testing, for example the test vectors given in the RFC, is hard to impossible from outside the crate

I'm not actually suggesting opaque-ke to provide a P256 implementation, though I don't mind if it does and I'm willing to provide the PR. But it seems that the current provided interface has to be improved either way to make implementations work.

daxpedda:
- Implements a generic holy the `Group` for both `PublicKey` and `PrivateKey`
- `Group::Scalar` now requires `Copy`
- `Group::scalar_as_bytes` now takes and returns a value instead of a reference
- `Group::hash_to_curve` now returns a `Result`
- expose `map_to_curve::expand_message_xmd`

There are quiet a bit of breaking changes, I didn't expect me doing this now, but it turned out that P256 is more important then we thought for our needs. I'm open to any suggestions.

Fixes #198.


src/keypair.rs
```diff
@@ -174,69 +169,131 @@ impl Deref for Key {
...
+    fn new(key: Key) -> Self {
+        Self {
+            key,
+            _g: PhantomData,
+        }
+    }
+}
+
+impl<G: Group> SizedBytes for PrivateKey<G> {
+    type Len = G::ScalarLen;
```
daxpedda, line 230: I'm not entirely sure if this is correct: the `PrivateKey` has `Group::ScalarLen` size and the `PublicKey` has `Group::ElemLen` size.

src/keypair.rs
```diff
@@ -174,69 +169,131 @@ impl Deref for Key {
 
 // Don't make it implement SizedBytes so that it's not constructible outside of this module.
 impl Key {
-    fn to_arr(&self) -> GenericArray<u8, KeyLen> {
+    fn to_arr<L: ArrayLength<u8>>(&self) -> GenericArray<u8, L> {
```
daxpedda, line 172: We could potentially store a `GenericArray` in `Key` instead of a `Vec`, now that we have a generic on both `PublicKey` and `PrivateKey`.

daxpedda:
Clippy is currently broken because of https://github.com/rust-lang/rust-clippy/issues/7422.

src/keypair.rs
kevinlewi: Yes!

src/keypair.rs
daxpedda: Done.

kevinlewi (approved):
Thanks, overall this is a great change!

Issue #199, opened 2021-07-06
Remote private key


daxpedda:
Currently there is no support for remote private keys in opaque-ke. Meaning that the use of HSMs or "vault storage"s to store the private key in `ServerSetup` isn't currently possible.

I am working towards an implementation and am planning to provide a PR to support this. Is my assumption correct that the only work the private key does is a DH key exchange, basically the remote key has to provide a replacement of this function?:
https://github.com/novifinancial/opaque-ke/blob/349329cdebf35bbb3767413ea8905aa4ababbf02/src/keypair.rs#L109-L118
I understand if opaque-ke doesn't want to implement support for this, but at least some API changes have to be made to allow external implementations to provide support.

kevinlewi:
Can you clarify what you mean by "remote private key" here? Do you mean that the HSM already has a designated private key, which you would like to use in OPAQUE instead of the one that that is generated from a `ServerSetup`?

If so, would it perhaps be simpler to just use that private key to encrypt the `ServerSetup` object, so that only the HSM can decrypt it, and then proceed with OPAQUE without having to worry about injecting this external HSM private key into the AKE for OPAQUE?

Hope that makes sense...

daxpedda:
> Can you clarify what you mean by "remote private key" here? Do you mean that the HSM already has a designated private key, which you would like to use in OPAQUE instead of the one that that is generated from a `ServerSetup`?

Really the idea is to never commit the private key to memory, therefore opaque-ke never get's access to the private key. Which then requires all necessary cryptographic functions to be executed by the device holding the private key, HSMs are built exactly for that purpose.

> If so, would it perhaps be simpler to just use that private key to encrypt the `ServerSetup` object, so that only the HSM can decrypt it, and then proceed with OPAQUE without having to worry about injecting this external HSM private key into the AKE for OPAQUE?

If the goal would be to never leak the private key of the server setup to the server, this isn't an option. This also allows to use the same keypair that is used for other purposes, like for example the HTTPS certificate.

kevinlewi:
I am not quite sure we are considering the same proposal (perhaps there is still a misunderstanding), let me try to be more concrete with what I mean. And sorry if this is too pedantic :)

So, typically an HSM has some private keypair (SK, PK), with the guarantee that SK never leaves the HSM. It seems that you want to execute OPAQUE within the HSM. When running ServerSetup, another keypair gets generated that is used only in OPAQUE: let's call it (OSK, OPK).

[...]

PR #212, opened 2021-07-16
Introduce customizable `PrivateKey`

daxpedda:
This is the current approach I took to tackle #199.

Using a dynamic object turned out not to be an option, mainly because I couldn't figure out serialization without breaking object safety.

~~Other approaches that I could think off that avoid having users specify a `PrivateKey` type:~~
- ~~Hiding the `PrivateKey` associated type somewhere else, for example in `Group` or `KeyExchange`~~
- ~~Introduce a default associated type `()` with the help of some hacks, because it isn't stable yet~~
- ~~Put the type in `ServerSetup` and pass it down to corresponding functions instead of putting it in `CipherSuite`~~

I decided to take the last approach, now the user-facing API almost doesn't change at all.

### DIFF
--- a/src/keypair.rs
+++ b/src/keypair.rs
@@ -7,7 +7,7 @@
 
 #![allow(unsafe_code)]
 
-use crate::errors::{InternalPakeError, PakeError, ProtocolError};
+use crate::errors::{InternalPakeError, ProtocolError};
 use crate::group::Group;
 #[cfg(test)]
 use generic_array::typenum::Unsigned;
@@ -286,7 +286,7 @@ pub trait SecretKey<G: Group>: Clone + Sized + Zeroize {
     fn serialize(&self) -> Vec<u8>;
 
     /// Deserialization from bytes
-    fn deserialize(input: &[u8]) -> Result<Self, ProtocolError>;
+    fn deserialize(input: &[u8]) -> Result<Self, InternalPakeError>;
 }
 
 impl<G: Group> SecretKey<G> for PrivateKey<G> {
@@ -308,8 +308,8 @@ impl<G: Group> SecretKey<G> for PrivateKey<G> {
         self.to_vec()
     }
 
-    fn deserialize(input: &[u8]) -> Result<Self, ProtocolError> {
-        PrivateKey::from_bytes(input).map_err(|_| PakeError::SerializationError.into())
+    fn deserialize(input: &[u8]) -> Result<Self, InternalPakeError> {
+        PrivateKey::from_bytes(input).map_err(InternalPakeError::from)
     }
 }
 
--- a/src/keypair.rs
+++ b/src/keypair.rs
@@ -280,6 +280,7 @@ impl<G: Group> SizedBytes for PrivateKey<G> {
 
 /// A trait specifying the requirements for a private key container
 pub trait SecretKey<G: Group>: Clone + Sized + Zeroize {
+    /// Custom error type that can be passed down to `InternalPakeError::Custom`
     type Error;
 
     /// Diffie-Hellman key exchange implementation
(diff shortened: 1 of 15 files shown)


src/keypair.rs
```diff
@@ -263,6 +274,45 @@ impl<G: Group> SizedBytes for PrivateKey<G> {
...
+    fn deserialize(input: &[u8]) -> Result<Self, ProtocolError>;
+}
+
+impl<G: Group> SecretKey<G> for PrivateKey<G> {
+    fn diffie_hellman(&self, pk: PublicKey<G>) -> Result<Vec<u8>, InternalPakeError> {
+        let pk_data = GenericArray::<u8, G::ElemLen>::from_slice(&pk.0[..]);
+        let point = G::from_element_slice(pk_data)?;
+        let secret_data = GenericArray::<u8, G::ScalarLen>::from_slice(&self.0[..]);
+        Ok(G::mult_by_slice(&point, secret_data).to_arr().to_vec())
+    }
```
daxpedda, line 298: I didn't realize this is all that DH is, is this correct? In any case, I'm about to test it soon.

src/keypair.rs
```diff
@@ -104,23 +113,25 @@ impl<G: Group> KeyPair<G> {
...
-        let point = G::from_element_slice(pk_data)?;
-        let secret_data = GenericArray::<u8, G::ScalarLen>::from_slice(&sk.0[..]);
-        Ok(G::mult_by_slice(&point, secret_data).to_arr().to_vec())
-    }
-
     /// Obtains a KeyPair from a slice representing the private key
-    pub fn from_private_key_slice(input: &[u8]) -> Result<Self, InternalPakeError> {
-        let sk = PrivateKey(Key(GenericArray::clone_from_slice(input)));
-        let pk = Self::public_from_private(&sk);
+    pub fn from_private_key_slice(input: &[u8]) -> Result<Self, ProtocolError> {
```
daxpedda, line 117: I believe this is the only user-facing API change so far.

src/keypair.rs
daxpedda: I confirmed this now.

daxpedda:
I managed to implement a custom error type to `SecretKey` without any breaking changes actually. I had to change a result type in the `KeyExchange` trait, but the trait was never exposed publicly, so no issues there.

The way I did that is that to add a `Custom` variant to `InternalPakeError`, with the default being `Infallible`, it turned out to be quiet messy because of conversions and what not, but it works. I'm open to any suggestions, I tried my best to keep it user friendly and not to big of a change.

daxpedda:
The tests are failing because the newest version of zeroize has a MSRV of 1.51.

kevinlewi:
This is fantastic, thanks for putting all of the effort into the change. As a follow-up, it would be great to add documentation to the lib.rs for how to use remote keys (probably under Advanced Usage). Creating https://github.com/novifinancial/opaque-ke/issues/214 to track this

kevinlewi (approved):

kevinlewi:
> The tests are failing because the newest version of zeroize has a MSRV of 1.51.

I'd like to understand better why this is the case... how come the newest version of zeroize is being pulled in here?

daxpedda:
If you don't specify otherwise, Cargo will use the caret requirement ([ref](https://doc.rust-lang.org/cargo/reference/specifying-dependencies.html#specifying-dependencies-from-cratesio)):
`zeroize = "1.1.1"` is really `zeroize = "^1.1.1"`
The caret requirement is described like this ([ref](https://doc.rust-lang.org/cargo/reference/specifying-dependencies.html#caret-requirements)):
> **Caret requirements** allow SemVer compatible updates to a specified version. An update is allowed if the new version number does not modify the left-most non-zero digit in the major, minor, patch grouping.

Basically what happened is that zeroize didn't consider bumping the MSRV a breaking change, a lot of crates do this. Do solve this issue here you can either specify a different requirement, most likely what we are looking for here is the [tilde requirement](https://doc.rust-lang.org/cargo/reference/specifying-dependencies.html#specifying-dependencies-from-cratesio), or bump the MSRV.

daxpedda:
I added further documentation under "Advanced usage" to deal with #214. I can add a full example later when we do more progress in Custodian to have actual support for remote keys or find a different library that does that already.

PR #213, opened 2021-07-20
P256 implementation

daxpedda:
I'm not a cryptographer, so I will need some guidance here on what to do, but it works as far as I've tested.
Any suggestions how or where I should put tests and further documentation?


src/group.rs
```diff
@@ -163,3 +163,74 @@ impl Group for RistrettoPoint {
         constant_time_eq::constant_time_eq(&self.to_arr(), &other.to_arr())
     }
 }
+
+#[cfg(feature = "p256")]
+impl Group for p256::ProjectivePoint {
```
daxpedda, line 168: I've used `ProjectivePoint`, but I don't know if that's actually correct.

src/group.rs
```diff
@@ -163,3 +163,74 @@ impl Group for RistrettoPoint {
...
+    ) -> Result<Self, InternalPakeError> {
+        use p256::elliptic_curve::sec1::FromEncodedPoint;
+
+        Self::from_encoded_point(&p256::EncodedPoint::from_secret_key(
+            &p256::SecretKey::new(p256::ScalarBytes::from_scalar(
+                &p256::Scalar::from_bytes_reduced(uniform_bytes),
+            )),
+            true,
+        ))
+        .ok_or(InternalPakeError::PointError)
```
daxpedda, line 219: We discussed this in #198, you mentioned we could use [`hash2field::hash_to_field`](https://docs.rs/hash2field/0.2.0/hash2field/fn.hash_to_field.html) here, I'm still very confused about this part.

As far as I understood, there are two parts, the first part is to use `expand_message`, that's currently done in `GroupWithMapToCurve::map_to_curve`.
I don't know what the second part does, something I don't understand, but apparently this is what `RistrettoPoint::from_uniform_bytes` takes care of. So in the case of P256 we will have to implement it manually or use the `hash2field` crate, it only has one function that does both though, so either we will have to change the interface or make `GroupWithMapToCurve::map_to_curve` no-op?

src/group.rs
daxpedda: As far as I understand this doesn't really matter, but it makes necessary math more accessible.

src/group.rs
daxpedda: `hash2field` sadly doesn't provide any of the utility we really need. In the and, after a lot of guidance from different people, the missing part was `map_to_curve`, which I had to implement by hand. The question of where to put the code is still open.

daxpedda:
MSRV is going to be an issue here, the last version of p256 to support 1.41 is [0.4.1](https://crates.io/crates/p256/0.4.1). We can make the `p256` crate feature have a different MSRV or increase the MSRV for the new version?

src/map_to_curve.rs
```diff
@@ -56,6 +56,35 @@ impl GroupWithMapToCurve for RistrettoPoint {
     }
 }
 
+#[cfg(feature = "p256")]
+impl GroupWithMapToCurve for p256_::ProjectivePoint {
+    const SUITE_ID: usize = 0x0003;
```
daxpedda, line 61: I'm not sure where I got this from, but looking at suite ids in the OPAQUE draft, we aren't following the suggested ones anyway. Is `0x0003` alright? I guess `0x0002` could be reserved for decaf448.

src/group.rs
```diff
@@ -163,3 +163,509 @@ impl Group for RistrettoPoint {
...
+        // `B: 0x5ac635d8aa3a93e7b3ebbd55769886bc651d06b0cc53b0f63bce3c3e27d2604b`
+        let b = BigInt::parse_bytes(
+            b"5ac635d8aa3a93e7b3ebbd55769886bc651d06b0cc53b0f63bce3c3e27d2604b",
+            16,
+        )
+        .unwrap();
+        // `L: 48`
+        const L: usize = 48;
+        // `Z: -10`
+        let z = BigInt::from(-10);
```
daxpedda, line 234: These could be const, which isn't supported in Rust at the moment. lazy-static or similar could be used, which is currently only a dev-dependency.

src/group.rs
```diff
@@ -163,3 +163,509 @@ impl Group for RistrettoPoint {
...
+/// Corresponds to the map_to_curve_simple_swu() function defined in
+/// <https://datatracker.ietf.org/doc/html/draft-irtf-cfrg-hash-to-curve-11#appendix-F.2>
+#[allow(clippy::many_single_char_names)]
+fn map_to_curve_simple_swu(
+    u: &num_bigint::BigInt,
+    a: &num_bigint::BigInt,
+    b: &num_bigint::BigInt,
+    p: &num_bigint::BigInt,
+    z: &num_bigint::BigInt,
+) -> (num_bigint::BigInt, num_bigint::BigInt) {
```
daxpedda, line 288: I made it like this to support re-use, is this useful?

src/group.rs
```diff
@@ -163,3 +163,509 @@ impl Group for RistrettoPoint {
...
+            number.pow_internal(&(self.0 - 2))
+        }
+    }
+
+    /// Finite field arithmetic
+    #[derive(Clone)]
+    struct FieldElement<'a> {
+        number: BigInt,
+        f: &'a Field<'a>,
+    }
```
daxpedda, line 325: I would have preferred not to implement something like this by hand, but I can't seem to find an appropriate library.

src/map_to_curve.rs
```diff
@@ -56,6 +56,35 @@ impl GroupWithMapToCurve for RistrettoPoint {
...
+
+        // https://datatracker.ietf.org/doc/html/draft-irtf-cfrg-hash-to-curve-11#section-3
+        // `hash_to_curve` calls `hash_to_field` with a `count` of `2`
+        // https://datatracker.ietf.org/doc/html/draft-irtf-cfrg-hash-to-curve-11#section-5.3
+        // `hash_to_field` calls `expand_message` with a `len_in_bytes` of `count * L`
+        let uniform_bytes = expand_message_xmd::<H>(msg, dst, 2 * L)?;
+
+        <Self as Group>::hash_to_curve(&GenericArray::clone_from_slice(&uniform_bytes[..]))
+            .map_err(ProtocolError::from)
+    }
```
daxpedda, line 76: This here seems to be the crux of the problem for me. Following the draft, I am trying to implement [`hash_to_curve`](https://datatracker.ietf.org/doc/html/draft-irtf-cfrg-hash-to-curve-11#section-3), which looks like this:
```
   Steps:
   1. u = hash_to_field(msg, 2)
   2. Q0 = map_to_curve(u[0])
   3. Q1 = map_to_curve(u[1])
   4. R = Q0 + Q1              # Point addition
   5. P = clear_cofactor(R)
   6. return P
```
But the current implementation was for ristretto255 only, which implemented [`hash_to_ristretto255`](https://datatracker.ietf.org/doc/html/draft-irtf-cfrg-hash-to-curve-11#appendix-B):
```
   Steps:
   1. uniform_bytes = expand_message(msg, DST, 64)
   2. P = ristretto255_map(uniform_bytes)
   3. return P
```
The way we named and ordered things doesn't seem to entirely align with the spec:
- `GroupWithMapToCurve` has a method `map_to_curve`
- `Group` has a method `hash_to_curve`
- `GroupWithMapToCurve::map_to_curve` calls `Group::hash_to_curve`
- Ristretto255:
  - Step 1 (`uniform_bytes = expand_message(msg, DST, 64)`): in `GroupWithMapToCurve::map_to_curve`
  - Step 2 & 3 (`ristretto255_map(uniform_bytes)`): in `Group::hash_to_curve`
- P-256:
  - Step 1 (`hash_to_field(msg, 2)`): split in `GroupWithMapToCurve::map_to_curve` and `Group::hash_to_curve`
  - Step 2 - 6 (`Q0 = map_to_curve(u[0])`): in `Group::hash_to_curve`

The reason I had to split up the P-256 implementation like this, is because I didn't want to continuously re-en/decode types between calls.

What is the reason it is like that? What is the point of `GroupWithMapToCurve`?

kevinlewi: Indeed, the naming and separation of Group vs. GroupWithMapToCurve no longer makes sense. (It was like this back when the spec was a much older version, but now it's just confusing.) I would be in favor of simply consolidating the two traits, since there no longer needs to be a separation.

src/group.rs
kevinlewi: This is an impressive endeavor! Perhaps it would be worth it to split out a bunch of this code into a separate file (like p256.rs)?

kevinlewi (approved):
Since everything is behind a feature, I'd be OK with landing this first (after potentially addressing some nits mentioned below)! Thanks for putting the effort into this.

src/map_to_curve.rs
daxpedda: Ah, alright, I was looking at OPAQUE suite ids, which is totally different.

kevinlewi:
> MSRV is going to be an issue here, the last version of p256 to support 1.41 is [0.4.1](https://crates.io/crates/p256/0.4.1). We can make the `p256` crate feature have a different MSRV or increase the MSRV for the new version?

Let's increase the MSRV for the new version.

daxpedda:
Except getting it to actually work and moving some things into sub-modules, I will do a couple of things in follow up PRs:
- increase MSRV
- merge `GroupWithMapToCurve` and `Group`
- potentially separate AKE and OPRF, which will need a separate discussion too

src/map_to_curve.rs
daxpedda: I will do this in a follow up PR.

src/map_to_curve.rs
```diff
@@ -56,6 +56,47 @@ impl GroupWithMapToCurve for RistrettoPoint {
...
+        )
+        .unwrap();
+        // `P256_XMD:SHA-256_SSWU_RO_` has an `L` of `48`
+        const L: usize = 48;
+        let uniform_bytes = expand_message_xmd::<H>(input, dst, L)?;
+        let bytes = BigUint::from_bytes_be(&uniform_bytes).mod_floor(&p);
+
+        Ok(p256_::Scalar::from_bytes_reduced(GenericArray::from_slice(
+            &bytes.to_bytes_be(),
+        )))
```
daxpedda, line 96: This is the offending piece of code right now, I can't get this to work. According to all our tests `expand_message_xmd` works correctly, so I can only believe that `mod_floor(p)` is nonsense, or `from_bytes_reduced` just doesn't work here. Asking some questions in RustCrypto I was informed that there is a missing function `from_bytes_wide` that might be what I'm looking for. After many hours I finally gave up and decided implementing this function myself goes pretty much over my head, any help would be amazing.

daxpedda: I tried to re-follow the steps I did when I tried to debug this yesterday and realized I forgot to mention, that I couldn't confirm that there isn't something wrong earlier, or to be more precise, inside `opaque::oprf_key_from_seed()`, which translates to `DeriveKeyPair` from the ORPF draft.

src/ciphersuite.rs
```diff
@@ -24,7 +23,7 @@ pub trait CipherSuite {
     /// an extension trait PasswordToCurve that allows some customization on
     /// how to hash a password to a curve point. See `group::Group` and
     /// `map_to_curve::GroupWithMapToCurve`.
-    type Group: GroupWithMapToCurve<UniformBytesLen = <Self::Hash as Digest>::OutputSize>;
```
daxpedda, line 27: I hope removing this bound is fine, I guess when we merge both trait's it gonna be unnecessary anyway.

src/tests/opaque_test_vectors.rs
```diff
@@ -506,14 +805,36 @@ fn parse_identifiers(
...
+    #[cfg(feature = "p256")]
+    {
+        test_registration_request::<P256Sha256NoSlowHash>(P256_TEST_VECTORS)?;
+        //test_registration_response::<P256Sha256NoSlowHash>(P256_TEST_VECTORS)?;
+        //test_registration_upload::<P256Sha256NoSlowHash>(P256_TEST_VECTORS)?;
+        test_ke1::<P256Sha256NoSlowHash>(P256_TEST_VECTORS)?;
+        //test_ke2::<P256Sha256NoSlowHash>(P256_TEST_VECTORS)?;
+        //test_ke3::<P256Sha256NoSlowHash>(P256_TEST_VECTORS)?;
+        //test_server_login_finish::<P256Sha256NoSlowHash>(P256_TEST_VECTORS)?;
+        //test_fake_vectors::<P256Sha256NoSlowHash>(P256_FAKE_TEST_VECTORS)?;
```
daxpedda, line 827: These tests should all be broken because of `hash_to_scalar` as noted above.

src/group.rs
daxpedda: I hope the way I split up the modules now is acceptable?

src/map_to_curve.rs
daxpedda: I fixed it, I was using the wrong variable, I used `P`, but I should have used `n`.

src/tests/opaque_test_vectors.rs
daxpedda: All tests work now.

src/group.rs
daxpedda: Done.

daxpedda:
This is ready to be merged, follow-up PRs are in the works.

PR #216, opened 2021-07-29
Add tests against intermediate values test vectors

daxpedda:
This test shouldn't fail, I don't know why it does.


daxpedda:
If this test is fixed and useful, I don't mind this being merged of course.

daxpedda:
As discussed, I implemented the rest of the intermediate value test vector tests, I left some `client_public_key` and `envelope` out because they are checked by `envelope` anyway.

I hope the additional types I had to add aren't too awful.

kevinlewi (approved):
This is awesome. Thanks!

PR #217, opened 2021-07-30
Increase MSRV to 1.51

daxpedda:
We discussed 1.53 actually, but it seems no dependency needs it, I also noticed that we don't document MSRV anywhere. Do you want me to add documentation? Is 1.51 alright or should I bump it to the newest right now, which is 1.54 actually?

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -17,7 +17,7 @@ jobs:
           - u32_backend
         toolchain:
           - stable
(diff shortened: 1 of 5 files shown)


kevinlewi:
Looks good, let's stick with 1.51 for now. I added documentation in the README and lib.rs for the MSRV. Thank you!

kevinlewi (approved):

PR #221, opened 2021-07-31
P-256 fix

daxpedda:
P-256 can in rare cases produce small numbers. `BigInt`, being a poor replacement for proper field elements, will produce an undesirably sized `Vec` on conversion.


kevinlewi (approved):
LGTM

PR #220, opened 2021-07-31
Seperate AKE from OPRF

daxpedda:
Based on #219. Makes cfrg/draft-irtf-cfrg-opaque#238 possible.

PR #222, opened 2021-08-03
Separate AKE from OPRF take 2

daxpedda:
As discussed, replaces #220.
I also added an implementation for X25519 and a test.

### DIFF
--- a/src/ciphersuite.rs
+++ b/src/ciphersuite.rs
@@ -11,6 +11,7 @@ use crate::{group::Group, hash::Hash, key_exchange::traits::KeyExchange, slow_ha
 /// * `Group`: a finite cyclic group along with a point representation, along
 ///   with an extension trait PasswordToCurve that allows some customization on
 ///   how to hash a password to a curve point. See `group::Group`.
+/// * `AkeGroup`: A `Group` used for the `KeyExchange`.
 /// * `KeyExchange`: The key exchange protocol to use in the login step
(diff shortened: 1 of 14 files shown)


src/ciphersuite.rs
```diff
@@ -19,8 +20,10 @@ pub trait CipherSuite {
     /// an extension trait PasswordToCurve that allows some customization on
     /// how to hash a password to a curve point. See `group::Group`.
     type Group: Group;
+    /// A `Group` used for the `KeyExchange`.
+    type AkeGroup: Group;
```
kevinlewi, line 24: Perhaps can we rename this type to KeGroup instead of AkeGroup?

And then we could also consider renaming the type on line 22 from "Group" to "OprfGroup"... though I am less confident about that one.

kevinlewi (approved):
This is pretty neat, thanks for doing this!

In lib.rs, we will also need documentation in the "Overview" section on the new type that needs to be specified by the consumer of the library for Ciphersuite. If you can add that to this PR, or create a tracking issue and I'll add it later, that would be great.

src/ciphersuite.rs
daxpedda: I think it's easier for the consumer if we rename to `OprfGroup`, even if the naming might not align to wording from the spec.

daxpedda:
I added the documentation to the "Overview", please feel free to add any improvements.

PR #223, opened 2021-08-06
Adding no_std support + alloc feature

kevinlewi:
Adds support for no_std for WebAssembly targets. This also introduces the `alloc` feature, not enabled by default.

### DIFF
--- a/src/errors.rs
+++ b/src/errors.rs
@@ -4,14 +4,11 @@
 // LICENSE file in the root directory of this source tree.
 
 //! A list of error types which are produced during an execution of the protocol
-use std::convert::Infallible;
-use std::error::Error;
-use std::fmt::Debug;
-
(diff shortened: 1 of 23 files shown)


src/errors.rs
```diff
@@ -4,14 +4,11 @@
 // LICENSE file in the root directory of this source tree.
 
 //! A list of error types which are produced during an execution of the protocol
-use std::convert::Infallible;
-use std::error::Error;
-use std::fmt::Debug;
-
-use displaydoc::Display;
+use core::convert::Infallible;
+use core::fmt::{Debug, Display};
```
daxpedda, line 8: We should leave the `Error` implementations, we but we should just hide it behind the `std` crate feature

kevinlewi:
Closing in favor of #225

PR #224, opened 2021-08-06
Updating dependencies for v1 and fixing zeroize tests

kevinlewi:


kevinlewi:
Closing this since it requires updating the MSRV (which I would like to avoid doing for v1)

PR #225, opened 2021-08-11
`no_std` support

daxpedda:
Alternative to #223.

As far as I can see, we only need `std` to implement `Error` and testing, otherwise it's completely out of the picture.
CI still needs to be adjusted.


daxpedda:
I wasn't sure what else to do with CI, just covering all the bases here I guess.

kevinlewi (approved):
Awesome, thanks! As discussed offline, it would be great to ensure that the dependencies are covered with no_std as well, in a future change.

PR #234, opened 2021-08-31
Add `SlowHash` configuration

### LINKED ISSUE #180: SlowHash configurability

daxpedda:
As discussed. I also upgraded the argon2 dependency to 0.3 along the way.

Fixes #180.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -22,7 +22,7 @@ std = ["curve25519-dalek/std", "getrandom", "rand/std", "rand/std_rng", "num-big
 serialize = ["serde", "base64", "generic-array/serde", "curve25519-dalek/serde"]
 
 [dependencies]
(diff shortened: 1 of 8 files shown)


kevinlewi:
The only nit I added/changed here is to change the generic type argument in the "Finish Parameters" to be for CipherSuite instead of SlowHash. The reason being that all other top-level APIs and structs also are generic in CipherSuite.

kevinlewi (approved):

PR #241, opened 2021-09-27
Constant time `cmov`

### LINKED ISSUE #239: Non-constant time `cmov` in P256
You should use something like what curve25519_dalek uses:
https://doc.dalek.rs/subtle/trait.ConditionallySelectable.html#method.conditional_assign
https://github.com/novifinancial/opaque-ke/blob/d1dfee9a6545d7dfddc6dd94f985f31b3221c54e/src/group/p256.rs#L327-L333

daxpedda:
@Sc00bz I basically copied what crypto-bigint does, but I don't bother converting to bigger sizes and just compare the `u8`s.

Fixes #239.

### DIFF
--- a/src/group/p256.rs
+++ b/src/group/p256.rs
@@ -164,6 +164,10 @@ impl Group for ProjectivePoint {
 
 /// Corresponds to the map_to_curve_simple_swu() function defined in
 /// <https://datatracker.ietf.org/doc/html/draft-irtf-cfrg-hash-to-curve-11#appendix-F.2>
+// `cmov`, `mod_floor` and `modpow` needs to be made constant-time, which
+// will be supported after crypto-bigint is no longer experimental. See
+// https://github.com/novifinancial/opaque-ke/issues/239 for more context.
+
 #[allow(clippy::many_single_char_names)]
 fn map_to_curve_simple_swu<N: ArrayLength<u8>>(
     u: &[u8],
@@ -327,9 +331,6 @@ fn map_to_curve_simple_swu<N: ArrayLength<u8>>(
         }
     }
 
-    // This needs to be made constant-time, which will be supported after
-    // crypto-bigint is no longer experimental. See
-    // https://github.com/novifinancial/opaque-ke/issues/239 for more context.
     fn cmov<'a>(x: &FieldElement<'a>, y: &FieldElement<'a>, b: bool) -> FieldElement<'a> {
         let f = x.f;
 
--- a/src/group/p256.rs
+++ b/src/group/p256.rs
@@ -26,6 +26,7 @@ use p256_::elliptic_curve::subtle::ConstantTimeEq;
 use p256_::elliptic_curve::Field;
 use p256_::{AffinePoint, EncodedPoint, ProjectivePoint};
 use rand::{CryptoRng, RngCore};
+use subtle::{Choice, ConditionallySelectable};
 
 // `L: 48`
 pub const L: usize = 48;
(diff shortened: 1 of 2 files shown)


Sc00bz:
You should use `to_bytes_le` and `from_bytes_le` instead of `to_signed_bytes_le` and `from_signed_bytes_le` because the current code breaks on negative numbers. Although you shouldn't have negative numbers. Also 256 bytes is excessive. Since these should fit in 32 bytes (33 bytes if you keep `to_signed_bytes_le`).

daxpedda:
I can't figure out why it would break on negative numbers, I did a small test converting with `to_signed_bytes_le` to `from_signed_bytes_le` on a negative number and it worked just fine.
Would you mind collaborating on why we can't encounter negative numbers? If that is the case, maybe we could already switch to `crypto_bigint::UInt`?

Sc00bz:
The code is zero filling. Take -1 then `to_signed_bytes_le` should return a one byte array of `0xff`. Then you copy that into a larger array and get `0xff, 0, 0, 0, ...`. Then call `from_signed_bytes_le` which will read it as 255.

> Would you mind collaborating on why we can't encounter negative numbers?

Everything should be done in a finite field thus all numbers are in the range 0 to p-1.

daxpedda:
Thanks a lot, I keep getting reminded I'm not a cryptographer :laughing:.
Now that I think about it there might be a way to switch to `crypto_bigint::UInt` after all.

Sc00bz:
The only thing you need to watch out for is subtraction and negation. Which are `(a + (p - b)) % p` and `(p - a) % p`. Wait **crypto**_bigint that's still under development: "[We recommend that it only be used in an experimental capacity for now.](https://docs.rs/crypto-bigint/0.2.2/crypto_bigint/#status)" But once that is finished then definitely use that as it will be constant time.

daxpedda:
Yeah, negation does happen indeed, specifically in the case of `cmov`. So I attempted some kind of constant-time check for the sign, @Sc00bz would you mind reviewing this again?

Sc00bz:
```
let sign = if u8::conditional_select(&0, &1, choice) == 0 {
```
Is the same as doing this:
```
let sign = if b {
```
Also this introduces a branch.

Oh I just looked up `mod_floor` and it returns `[0, p)` if `p > 0`. So you do not need to worry about the sign. You should remove that if statement and use `Sign::Plus`. You should call `mod_floor` on `x` and `y` before converting them to byte arrays. Also note `mod_floor`, `modpow`, and `x[..x_bytes.len()].copy_from_slice(&x_bytes);` are not constant time but they are not currently easy fixes. Basically you'd need to implement your own bigint. The only "easy fix" is use `crypto_bigint` when it's finished.

daxpedda:
I can't ignore the sign, `cmov` is specifically called at the end of the algorithm to switch the sign:
https://github.com/novifinancial/opaque-ke/blob/65a0c2f98d592bd820a14900266cf909f7add949/src/group/p256.rs#L388-L389
@kevinlewi I suggest we mark the p256 implementation as experimental and merge this as a best effort for now until crypto-bigint is ready.

kevinlewi:
I added a warning for p256 being experimental because of this. Thanks @Sc00bz for providing the feedback and originally raising the issue, and thanks @daxpedda for handling it!

Sc00bz:
If you do this then the use of [mod_floor](https://docs.rs/num-bigint/0.3.1/num_bigint/struct.BigInt.html#method.modpow) fixes the signing issue.

```
     fn cmov<'a>(x: &FieldElement<'a>, y: &FieldElement<'a>, b: bool) -> FieldElement<'a> {
         let f = x.f;

-        let (x_sign, x_bytes) = x.number.to_bytes_le();
+        let x_bytes = x.number.mod_floor(f.0).to_bytes_le().1;
         let mut x = [0; 32];
         x[..x_bytes.len()].copy_from_slice(&x_bytes);

-        let (y_sign, y_bytes) = y.number.to_bytes_le();
+        let y_bytes = y.number.mod_floor(f.0).to_bytes_le().1;
         let mut y = [0; 32];
         y[..y_bytes.len()].copy_from_slice(&y_bytes);

         let mut bytes = [0; 32];

         let choice = Choice::from(u8::from(b));

-        let sign = if u8::conditional_select(&0, &1, choice) == 0 {
-            x_sign
-        } else {
-            x.clone()
-            y_sign
-        };
-
         for ((byte, x), y) in bytes.iter_mut().zip(x).zip(y) {
             *byte = u8::conditional_select(&x, &y, choice);
         }

         FieldElement {
             f,
-            number: BigInt::from_bytes_le(sign, &bytes),
+            number: BigInt::from_bytes_le(Sign::Plus, &bytes),
         }
     }
```

daxpedda:
I was a bit confused why this actually works, but I realized that what @Sc00bz is proposing already happens anyway.

The negation is implemented on `FieldElement` with `Neg`, which already does `mod_floor`. So yes, the number will always be positive. Also realized that we don't have to worry about subtraction, which I just removed anyway, which in any case, was also using `mod_floor`.

Sc00bz:
I was just copying this. If `number` is always [0, p) then you can remove this `mod_floor`:

https://github.com/novifinancial/opaque-ke/blob/f08abd7b0f81e8f6a984e009cd9a836eb20f83f3/src/group/p256.rs#L322-L324

> @Sc00bz, would you mind making sure I'm not talking nonsense here?

OK I'm pretty sure you are correct.

kevinlewi:
Thanks everyone!

PR #244, opened 2021-09-30
P256 improvements

daxpedda:
Porting some improvements from VOPRF, I guess this is unnecessary in the long run when we include VOPRF as a dependency?

See novifinancial/voprf#18.


kevinlewi:
Yes, that is true.

kevinlewi (approved):

PR #248, opened 2021-10-17
Using voprf as a dependency

kevinlewi:
This will remove the OPRF implementation in this library and instead rely on the voprf crate at https://github.com/novifinancial/voprf .

Still in draft mode, with the remaning TODOs:
- Need voprf new version to be published with extra functions exposed in the "danger" feature
- Need to reinstate the x25519 group implementation


src/impls.rs
```diff
@@ -105,64 +105,39 @@ macro_rules! impl_clone_for {
...
             {
+                use serde::ser::Error;
+
                 if serializer.is_human_readable() {
-                    serializer.serialize_str(&base64::encode(&self.serialize()))
+                    serializer
+                        .serialize_str(&base64::encode(&self.serialize().map_err(Error::custom)?))
                 } else {
-                    serializer.serialize_bytes(&self.serialize())
+                    serializer.serialize_bytes(&self.serialize().map_err(Error::custom)?[..])
```
daxpedda, line 120:
```suggestion
                    serializer.serialize_bytes(&self.serialize().map_err(Error::custom)?)
```

src/key_exchange/group/p256.rs
```diff
@@ -0,0 +1,45 @@
...
+    fn public_key(sk: &GenericArray<u8, Self::SkLen>) -> Self {
+        Self::generator() * p256_::Scalar::from_bytes_reduced(sk)
+    }
+
+    fn to_arr(&self) -> GenericArray<u8, Self::PkLen> {
+        use p256_::elliptic_curve::sec1::ToEncodedPoint;
+
+        let mut bytes = self.to_affine().to_encoded_point(true).as_bytes().to_vec();
+        bytes.resize(33, 0);
+        *GenericArray::from_slice(&bytes)
```
daxpedda, line 39:
```suggestion
        let bytes = self.to_affine().to_encoded_point(true);
        let bytes = bytes.as_bytes();
        let mut result = GenericArray::default();
        result[..bytes.len()].copy_from_slice(bytes);
        result
```

Cargo.toml
```diff
@@ -14,12 +14,12 @@ resolver = "2"
 [features]
 default = ["u64_backend", "serialize"]
 slow-hash = ["argon2"]
-p256 = ["num-bigint", "num-integer", "num-traits", "once_cell", "p256_"]
+p256 = ["num-bigint", "num-integer", "num-traits", "once_cell", "p256_", "voprf/p256"]
```
daxpedda, line 17: The num crates and once_cell can be removed now.

daxpedda (changes requested):
Really small changes here and there, LGTM. I will port improvements similar to what I have done in voprf in separate PR's.

A thought: we could (should?) probably move to higher-level libraries for `KeGroup`. [x22519-dalek](https://crates.io/crates/x25519-dalek) for X22519 and [`ecdh`](https://docs.rs/elliptic-curve/0.10.6/elliptic_curve/ecdh/index.html) for P256. I also think these can be handled in different PR's.

PR #249, opened 2021-10-24
Fix wrong de-serialization of `Ke2Message`

daxpedda:
The wrong `Group` was used here during de-serialization. I wanted to add a test, but was unsure where to add it, any suggestions?


daxpedda:
Sounds good, this bug was indeed fixed by #248.

PR #250, opened 2021-10-26
General improvements

daxpedda:
General improvements. Many changes are ported from novifinancial/voprf#24, novifinancial/voprf#26, novifinancial/voprf#34 and novifinancial/voprf#35.

Noteworthy:
- Fix wrong de-serialization on `Ke1State`.
- Remove unnecessary `Result` from `KeyPair::generate_random`.
- Fix de-serialization when using owned bytes, like with `bincode::deserialize_from` (see novifinancial/voprf#33).
- Replace macros with [derive_where](https://crates.io/crates/derive-where) fixing the `Debug` implementation and missing wrong bounds.
- Implement `Ord` and `PartialOrd` on high level API items.
- Add `rust-version` to `Cargo.toml`.
- Remove lots of allocations.
- Remove human-readable base64 serialization from Serde implemenetation.
- Make curve25519-dalek backend optional.
- Rename `serialize` to `serde`.
- Switch `KeGroup` implementations to higher-level libraries:
  - p256 uses `elliptic_curve::PublicKey` with `p256::NistP256` - or `p256::PublicKey`.
  - X25519 uses `x25519_dalek::PublicKey`. This also surfaced a bug, it was missing clamping keys.


daxpedda:
I'm attempting to fix our MSRV issues here: yoshuawuyts/fd-lock#17.

daxpedda:
As mentioned on Slack, in my attempt to reduce allocations, I will introduce very ugly `where` bounds. I considered it a non-issue because upgrading to digest 0.10 will introduce even worse ones. See RustCrypto/KDFs#54.

daxpedda:
I managed to avoid almost all `where` clauses and limit them only to `serialize` functions.

daxpedda:
@kevinlewi now that I switched `KeGroup` to high-level-implementations, I was thinking of making separate features for re-exporting P-256 `OprfGroup` and `KeGroup`, as `OprfGroup` is experimental and currently requires `alloc` and a lot of dependencies. But P-256 is safe (I believe!) and doesn't require `alloc`. WDYT? `p256-ke` and `p256-oprf`?

src/key_exchange/group/p256.rs
```diff
@@ -9,39 +9,39 @@ use super::KeGroup;
...
     fn random_sk<R: RngCore + CryptoRng>(rng: &mut R) -> GenericArray<u8, Self::SkLen> {
-        use p256_::elliptic_curve::Field;
-
-        p256_::Scalar::random(rng).into()
+        SecretKey::<NistP256>::random(rng).to_bytes()
     }
 
     fn public_key(sk: &GenericArray<u8, Self::SkLen>) -> Self {
-        Self::generator() * p256_::Scalar::from_bytes_reduced(sk)
+        SecretKey::<NistP256>::from_bytes(sk).unwrap().public_key()
```
daxpedda, line 31: These `unwrap`s have to be removed.

My idea was to use the appropriate `SecretKey` type instead of storing `GenericArray`s everywhere. I believe this would be better anyway, because conversion isn't free (it's very cheap though) and any checking is now guaranteed in the type and can now be handled during serialization.

kevinlewi:
> @kevinlewi now that I switched `KeGroup` to high-level-implementations, I was thinking of making separate features for re-exporting P-256 `OprfGroup` and `KeGroup`, as `OprfGroup` is experimental and currently requires `alloc` and a lot of dependencies. But P-256 is safe (I believe!) and doesn't require `alloc`. WDYT? `p256-ke` and `p256-oprf`?

I think we should avoid splitting out into two features, `p256-ke` and `p256-oprf` -- this should no longer be needed given your changes to `voprf` which make it not require `alloc`, right?

daxpedda:
The issue is that our implementation of P-256 in `voprf` uses `num-bigint`, which requires `alloc`. The implementation is experimental anyway and to be honest shouldn't really be living in voprf.

When RustCrypto/elliptic-curves#495 is done, we won't need our custom implementation of P-256 in `voprf` and my suggestion would be moot anyway. So any feature split is only temporary.

I agree that this is not a nice solution, I see a couple of paths here we could take:
1. Leave everything as it and hope RustCrypto/elliptic-curves#495 gets resolved sometime soon.
2. Split it out into two different features (my suggestion above) and remove that when RustCrypto/elliptic-curves#495 is resolved.
3. Don't actually enable the `p256` feature of `voprf` through `opaque-ke`, users who want to use the experimental P-256 implementation in `voprf` have to manually enable it.

I'm leaning on solution 3. as a viable alternative to solve the issue at least temporarily.

daxpedda:
I think this PR is enough for now, the issue pointed above can be fixed separately.

kevinlewi:
Agreed, thanks for laying that out clearly. I think solution #3 is the way to go at the moment.

kevinlewi (approved):

PR #261, opened 2022-01-05
Group revamp

### LINKED ISSUE #254: `KeGroup` trait implementation
Currently `KeGroup` is implemented on what represents the public key of a curve, the `RistrettoPoint` for ristretto255, `ProjectivePoint` for P-256 and `MontgomaryPoint` for X25519. When I switched to higher-level implementation for #250 I realized that we potentially don't even need a custom trait anymore.

[`elliptic_curve::PublicKey`](https://docs.rs/elliptic-curve/latest/elliptic_curve/struct.PublicKey.html) requires a type that implements [`Curve`](https://docs.rs/elliptic-curve/latest/elliptic_curve/trait.Curve.html) and [`ProjectiveArithmetic`](https://docs.rs/elliptic-curve/latest/elliptic_curve/trait.ProjectiveArithmetic.html) for example.

We can't actually use those because the current ristretto255 implementation we use doesn't implement any of these traits. So instead I'm proposing the following:
- Change `KeGroup` to not be implemented on the public key, but on a trait that is supposed to implement `Curve` and `ProjectiveArithmetic`. We introduce two associated types for a public and secret key.
- We can't provide a blanket implement of `KeGroup` for anything that satisfies the right trait, because we lack specialization to also provide a custom implementation for `RistrettoPoint`, but we can provide a simple template as an example.
- In the future, when curve25519-dalek implements the necessary traits, we can do a blanket implementation and basically remove all curve dependencies and associated crate features. Unless Rust stabilizes specialization first, lol.

This idea can also be implemented for voprf.
This is a proposal, I am planning to do the implementation as soon as this is approved.

### LINKED ISSUE #256: Remove unwrap from public_key() function
These `unwrap`s have to be removed.

My idea was to use the appropriate `SecretKey` type instead of storing `GenericArray`s everywhere. I believe this would be better anyway, because conversion isn't free (it's very cheap though) and any checking is now guaranteed in the type and can now be handled during serialization and creation.

_Originally posted by @daxpedda in https://github.com/novifinancial/opaque-ke/pull/250#r773146032_

### LINKED ISSUE #258: `DeriveAuthKeyPair` implementation uses the wrong type
We are currently using the `OprfGroup` to use the `hash_to_field` function for `DeriveAuthKeyPair`. Which is incorrect, we need to use the `KeGroup`. An issue I noticed while working on this, is that there is no way to infallibly create a secret and public key with `hash_to_field`.

The only workaround I see here is just to keep looping until RNG yields a non-zero scalar.

See cfrg/draft-irtf-cfrg-opaque#298.

### LINKED ISSUE #267: Version 0.6.1 (draft-irtf-cfrg-opaque-05) support for p256/SHA256
Hi All

Is it possible to make somehow version 0.6.1 implementing draft-irtf-cfrg-opaque-05 running for p256/SHA256 something like 
main [draft-irtf-cfrg-opaque-07]
  type OprfGroup = p256_::ProjectivePoint;
    type KeGroup = p256_::PublicKey;
    type KeyExchange = opaque_ke::key_exchange::tripledh::TripleDH;
    type Hash = sha2::Sha256;
    type SlowHash = opaque_ke::slow_hash::NoOpHash;
or version 0.6.1 is only for Group = curve25519_dalek::ristretto::RistrettoPoint/Hash = sha2::Sha512

daxpedda:
~~The `Zeroize` implementation and testing got out of hand here. I will improve on that later. See RustCrypto/utils#652, RustCrypto/utils#699 and RustCrypto/utils#700 on progress how we can improve on that.~~

- Move de/serializing checks for keys to the `KeGroup` implementation ~~(X25519 still outstanding)~~.
- Removed separate `Hash` associated type in `CipherSuite` and now the one from `OprfGroup` is used.
- Removed P-256 implementation and replaced with generic one over RustCrypto traits.
- Use `ZeroizeOnDrop` trait where appropriate.
- Removed `Key` type and instead use type provided by `KeGroup`.
- Change back permissions for all files to `644` (probably messed up by me because I was using a MacBook).
- Removed documented experimental warning for P-256 :tada:.
- Consolidated some CI runs and added more feature combinations to test.
- Remove generic from `SlowHash` and move it to the method instead.
- Updated MSRV to 1.57 because of dependencies on `crypto-bigint`.
- Fix `DeriveKeyPair` implementation using the wrong group for `HashToField`.

Issues introduced :sweat_smile::
- Some places `ZeroizeOnDrop` had to be written by hand because the derive proc-macro doesn't support `GenericArray`.
- Can't update to `curve25519-dalek` v4.0.0-pre.2 because it constrained `zeroize` to `<=1.4`. See https://github.com/novifinancial/voprf/pull/64.
- Had to pull in `curve25519-dalek` v3 optionally for `x25519-dalek` because the newest version doesn't use v4. See https://github.com/dalek-cryptography/x25519-dalek/pull/87.

Fixes #254.
Fixes #256.
Fixes #258.
Fixes #267.


daxpedda:
This is currently broken for a couple of reasons:
- The new `ZeroizeOnDrop` derive doesn't work for `GenericArray` (currently just implementing it by hand).
- `derive_where` doesn't support `cfg` attributes on fields.
- I didn't adjust tests yet. I will document all the changes I made here in the OP when I have working progress.

Will figure out how to fix/report this upstream.

daxpedda:
@kevinlewi this should be ready to review. I will keep in draft because I'm going to push some more code-improvements.

kevinlewi (approved):
Excellent, this was a big one, thanks so much!!

Should I wait to merge for the general improvements you wanted to add, or go ahead and merge now?

daxpedda:
I can make the general improvements as a follow-up, so go ahead!

daxpedda:
Pushed the improvements I had here for now, feel free to merge (or tell me to stop :smile:) anytime.

Issue #264, opened 2022-01-24
Can't compile 0.6.0 version referenced in opaque-wasm, can you please help?


jaroslavsaxa:
trying to build opaque-wasm which refers opaque-ke 0.6.0
following error:

Compiling opaque-ke v0.6.0

--> opaque-ke-0.6.0/src/envelope.rs:64:17
   |
64 | #[derive(Clone, Copy, Debug, Eq, Hash, PartialEq, Zeroize)]
   |                 ^^^^ Copy not allowed on types with destructors
   |
   = note: this error originates in the derive macro `Copy` (in Nightly builds, run with -Z macro-backtrace for more info)

For more information about this error, try `rustc --explain E0184`.
error: could not compile `opaque-ke` due to previous error

rust version: 1.58.1

kevinlewi:
See https://github.com/novifinancial/opaque-ke/pull/237 (and https://github.com/iqlusioninc/crates/issues/876)

This fix is to essentially move off of zeroize v1.1... which was pinned opaque-ke v0.6.0 unfortunately. Would it be possible to override the zeroize dependency to beyond v1.1 for now, until we release a new version of opaque-ke (and opaque-wasm also updates)?

daxpedda:
I think the issue here is that @jaroslavsaxa is actually using the new `zeroize`/`zeroize_derive` version, which is fixed and actually implements `Drop`, the compilation then fails because `Copy` is implemented on our type, which isn't allowed.

Further, all versions of `zeroize_derive`, the package that actually had the issue, before that fix were yanked, so `cargo update -p zeroize_derive --precise 1.1.0` doesn't work.

@kevinlewi I think an easy fix would be to backport this fix and release a `0.6.1`?

daxpedda:
@kevinlewi I made a fix here: https://github.com/novifinancial/opaque-ke/compare/v0.6.0...khonsulabs:v0.6.1?expand=1
If you make a branch based on v0.6.0 I can PR against it.

kevinlewi:
I discussed offline (through email) with @jaroslavsaxa -- it seems like his problems were addressed since he could modify a local copy of opaque-ke for his purposes... so we can probably close this and forego doing a new release... for now.

jaroslavsaxa:
Hi Friends, will be version 0.6.1 published to crates.io so we can refererence it in opaque-wasm?

kevinlewi:
[...]

PR #266, opened 2022-01-25
Backport `zeroize` fix

daxpedda:
As discussed in #264.


daxpedda:
CI currently breaks of criterion, there is no way to really fix this. Considering it only applies to the MSRV, I consider this fine, MSRV still builds, criterion is a `dev-dependency`.

kevinlewi:
Hmm, not a huge fan of the failing CI. Perhaps we should just update the MSRV and release a 0.7.0?

daxpedda:
I don't think this is necessary, I can fix the CI by simply removing the benchmark and therefore the reliance on criterion.

Alternatively I can make another crate (making this a workspace) and put non-MSRV tests there, but I don't consider this worth the effort.

kevinlewi:
Ah I see! Yes, please go ahead and remove the benchmark (for the sake of fixing CI), plus the criterion dependency. Will be happy to merge and push a new release afterwards!

kevinlewi (approved):

daxpedda:
After removing criterion I encountered another issue with MSRV: `rustyline`. I didn't want to remove the examples, so continuing to follow down that path, `nix` is using `bitflags` v1, which breaks MSRV. So the latest version of `rustyline` using a version of `nix` that uses an earlier version of `bitflags` then v1, was `rustyline` v1, so I had to downgrade to that.

Turns out `proptest` is also using bitflags` v1, the earliest version of `proptest` not using `bitflags` v1 is v0.3, but that breaks everything ... looking into it.

daxpedda:
A lot of downgrading, @kevinlewi should be good to go now.

kevinlewi (approved):
Thanks!

PR #268, opened 2022-02-25
General Improvements

daxpedda:
- Require `Copy` for both public and private key and switch to points instead of using `PublicKey`-like types.
- Require `Zeroize` for both public and private key instead of only using `KeGroup::zeroize_sk_on_drop()` on the private key.
- Fixed some missing checks in the ristretto255 implementation for zero scalars and identity elements during deserialization.
- Base the X25519 implementation on `curve25519-dalek` instead of `x25519-dalek` to support `Zeroize` and drop two dependencies.
- Use `derive_where(ZeroizeOnDrop)` instead of manual implementations.
- Update `argon2` to v0.4.
- Remove unnecessary `getrandom` and `constant_time_eq` dependency.
- Update `derive-where` to v1.0.0-rc.3
- Remove unnecessary `base64`, `lazy_static` and `sha2` dev-dependencies.
- No multiple versions of the same dependency present anymore 🎉 (`clippy::multiple_crate_versions`).
- Rename `ristretto255_*` and `x25519_*` crate features to `ristretto255-*` and `x25519-*`.
- Remove unnecessary dependency feature activation of `rand/std`, `rand/std_rng` and `voprf/std`.
- Fix possibility to produce zero scalars through `KeGroup::hash_to_scalar` (this is handled by the newest OPAQUE draft and the check added here should be removed).
- Rename `CipherSuite::OprfGroup` to `CipherSuite::OprfCs`.
- Rename `TripleDH` to `TripleDh`.
- Remove `slow-hash` crate feature, `argon2` should be used instead.
- Rename `SlowHash` to `Ksf` (**K**ey **S**tretching **F**unction, as it is called in the spec).
- Remove custom `De/Serialize` implementation for most types :tada: (this removes the macro and uses `derive` instead).


src/key_exchange/group/ristretto255.rs
```diff
@@ -63,7 +62,7 @@ impl KeGroup for Ristretto255 {
                 }
             };
 
-            if scalar != Scalar::zero() && scalar.is_canonical() {
```
daxpedda, line 66: `Scalar::random()` uses `Scalar::from_bytes_mod_order_wide()` (which we used before), which should do proper clamping, so this check isn't needed. Is this correct?

kevinlewi: Agreed

kevinlewi (approved):
Thanks! Is this still meant to be in draft mode?

kevinlewi:
It would be nice to be able to sync the implementation with the latest draft and its changes (https://datatracker.ietf.org/doc/html/draft-irtf-cfrg-opaque-08.html). I'd be happy to work on that, but want to avoid any potential merge conflicts, since I will be pulling from the main branch. If you think it would be better to land these changes now, let me know and I can merge before starting my work.

Issue #276, opened 2022-06-21
Server registration finish without username


jonahbeckford:
This issue is in regards to:

https://github.com/novifinancial/opaque-ke/blob/5a0aef607ccda21ae8a76c687de801632dff2574/examples/simple_login.rs#L94-L99

The `ServerRegistration` has the password file but not the username (`credential_identifier`).  If you expose two server REST methods `/ServerRegistrationStart` and `/ServerRegistrationFinish`, your `/ServerRegistrationFinish` would have to authenticate the username some other way.

I may be missing something in the API, but simple_login.rs seems to assume that you will have a single connection for all the registration messages. A typical REST implementation would violate that assumption. If so, a documentation comment will help users not to forget this part of the authentication.

Thanks. Nice library!

daxpedda:
If I understand correctly, I think there is a misunderstanding here. [`ServerRegistration::finish`](https://docs.rs/opaque-ke/2.0.0-pre.2/opaque_ke/struct.ServerRegistration.html#method.finish) doesn't require any state data, it also doesn't require `credential_identifier`.

I assume, the problem was that you were using the `credential_identifier` as an identifier in your database to save the resulting [`ServerRegistration`](https://docs.rs/opaque-ke/2.0.0-pre.2/opaque_ke/struct.ServerRegistration.html). As far as I am concerned this has nothing to do with the OPAQUE protocol and is an implementation detail for the user to figure out.

Did I understand this correctly? Please feel free to elaborate on any point I might have misunderstood there!

jonahbeckford:
> If I understand correctly, I think there is a misunderstanding here. [`ServerRegistration::finish`](https://docs.rs/opaque-ke/2.0.0-pre.2/opaque_ke/struct.ServerRegistration.html#method.finish) doesn't require any state data, it also doesn't require `credential_identifier`.

Agreed.

> I assume, the problem was that you were using the `credential_identifier` as an identifier in your database to save the resulting [`ServerRegistration`](https://docs.rs/opaque-ke/2.0.0-pre.2/opaque_ke/struct.ServerRegistration.html). As far as I am concerned this has nothing to do with the OPAQUE protocol and is an implementation detail for the user to figure out.

I haven't saved it as a database identifier yet; just doing a sanity check of how it is meant to be used.

---

Together the following lib.rs documentation and the parameter name `credential_identifier` create an impression that the OPAQUE protocol will authenticate the `credential_identifier`.

[...]

PR #277, opened 2022-06-23
Document responsibility for server-side id. #276

jonahbeckford:
This is a passive warning to the server implementor to not blindly trust the client to send the same identifier in both `RegistrationRequest` and `RegistrationUpload`.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -89,9 +89,9 @@
 //! key-value database, where the keys consist of these server-side identifiers
 //! for each client, and the values consist of their corresponding password
 //! files, to be retrieved upon future login attempts made by the client.
-//! It is your responsibility to ensure that the identifier given by the
-//! client in the initial [RegistrationRequest] is the database key you use
-//! in the final [RegistrationUpload] step.
+//! It is your responsibility to ensure that the identifier used to form the
+//! initial [RegistrationRequest] matches the database key used in the final
+//! [RegistrationUpload] step.
 //!
 //! Note that the [RegistrationUpload] message contains sensitive information
 //! (about as sensitive as a hash of the password), and hence should be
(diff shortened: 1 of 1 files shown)


src/lib.rs
```diff
@@ -89,6 +89,9 @@
 //! key-value database, where the keys consist of these server-side identifiers
 //! for each client, and the values consist of their corresponding password
 //! files, to be retrieved upon future login attempts made by the client.
+//! It is your responsibility to ensure that the identifier given by the
+//! client in the initial [RegistrationRequest] is the database key you use
+//! in the final [RegistrationUpload] step.
```
daxpedda, line 94:
```suggestion
//! It is your responsibility to ensure that the identifier used in the initial
//! [RegistrationRequest], typically supplied by the client, is the same as the
//! database key used in the final [RegistrationUpload] step.
```
For the purpose of understanding this, I wanted to highlight the crux of the problem here, that in a typical scenario, the `credential_identifier` is supplied by the user, just to make sure not to give the impression that it's required.

It might be useful to mention REST API here, or lack of state? Maybe it's overkill.
Also my English is not the best, so I'm going to refrain from ruining this great PR even more.

daxpedda:
CC #276.

src/lib.rs
kevinlewi: This looks good!

The only edit I nits I would propose are to replace:
"used in the initial [RegistrationRequest]" -> "used to form the initial [RegistrationRequest]"
"is the same as" -> "matches"

And I think we don't need to mention REST API / statelessness here -- I'm happy with leaving it just at this additional sentence.

kevinlewi (changes requested):

src/lib.rs
jonahbeckford: Did the copy-editing. Thanks!

kevinlewi (approved):
LGTM -- will give @daxpedda a chance to view the text before merging.

kevinlewi:
Thanks for bringing up the issue and making the PR, @jonahbeckford!

PR #282, opened 2022-07-20
Argon2 Example

zer0x64:
As discussed here:
https://github.com/novifinancial/opaque-ke/issues/281

This PR aims to update the `simple_login` example to add an example on how to use Argon2 as the client-side KDF, as recommended. The `Argon2` struct requires a lifetime, which makes the empty struct pattern somewhat unintuitive without an example and requires a `PhantomData` to work.

I'm opening as a draft for now because:
1. I'd like to also make an example on how to specify the derivation params instead on relying on RustCrypto's default, but I'm not sure doing it in the `simple_login` example is the best idea since it's supposed to be simple.
2. @daxpedda also suggested adding a custom KDF example. I think copying the `simple_login` file but implementing `Ksf` on, let's say, scrypt would be a good idea for that? If you expect to eventually first-party support `scrypt` we could go with something more obscure like PBKDF2-BLAKE2/3 instead so the example stays relevant.


Cargo.toml
```diff
@@ -55,6 +55,7 @@ zeroize = { version = "1.5", features = ["zeroize_derive"] }
 getrandom = { version = "0.2", features = ["js"], optional = true }
 
 [dev-dependencies]
+argon2 = { version = "0.4", default-features = false, features = ["alloc"] }
```
daxpedda, line 58:
```suggestion
```
Examples have access to all dependencies anyway, so this isn't necessary.

daxpedda (changes requested):
> The `Argon2` struct requires a lifetime, which makes the empty struct pattern somewhat unintuitive without an example and requires a `PhantomData` to work.

A secret isn't used in this case, so using a `'static` lifetime avoids the workaround with a `PhantomData`.
> 1. I'd like to also make an example on how to specify the derivation params instead on relying on RustCrypto's default, but I'm not sure doing it in the `simple_login` example is the best idea since it's supposed to be simple.

We can add this to the example with a custom KSF.


> 2\. @daxpedda also suggested adding a custom KDF example. I think copying the `simple_login` file but implementing `Ksf` on, let's say, scrypt would be a good idea for that? If you expect to eventually first-party support `scrypt` we could go with something more obscure like PBKDF2-BLAKE2/3 instead so the example stays relevant.

I think scrypt is appropriate considering it's also suggested by the draft. I'm happy with whatever tough. This can also be added in a separate PR if desired.

zer0x64:
So I made the changes you asked for in the PR. A few notes:

1. I made sure that using `'static` for Argon2 does not break even if you use custom parameters. While you probably still need the `PhantomData` workaround if using a secret key, I can't tell how much of a niche use case that would be (I'm thinking about a case where you'd want to authenticate using both your password and some device key or keyfile).

2. For the custom `Ksf` example, I copied the entire `simple_login` example and modified it. I also included custom derivation parameters. I'm not sure if that's too overkill, but because of custom parameters we basically need the full login process, so I'm not sure how we could have a shorter "runnable" example.

3. I think maybe we shouldn't use `Default` for the Ciphersuite because it conflicts with `core::default::Default`. I'd suggest renaming it `DefaultCiphersuite` or something like that while I'm at it.

examples/custom_ksf.rs
```diff
@@ -0,0 +1,286 @@
...
+// The Ksf trait must be implemented to be used in the ciphersuite.
+impl opaque_ke::ksf::Ksf for CustomKsf {
+    fn hash<L: generic_array::ArrayLength<u8>>(
+        &self,
+        input: GenericArray<u8, L>,
+    ) -> Result<GenericArray<u8, L>, opaque_ke::errors::InternalError> {
+        let mut output = GenericArray::<u8, L>::default();
+        scrypt::scrypt(&input, &self.salt, &self.params, &mut output).map_err(
+            |_: scrypt::errors::InvalidOutputLen| opaque_ke::errors::InternalError::KsfError,
+        )?;
```
daxpedda, line 45:
```suggestion
        let mut output = GenericArray::default();
        scrypt::scrypt(&input, &self.salt, &self.params, &mut output).map_err(
            |_| opaque_ke::errors::InternalError::KsfError,
        )?;
```
Is Rust really incapable of implicitly determining the type here?

examples/custom_ksf.rs
```diff
@@ -0,0 +1,286 @@
...
+    ServerLoginStartParameters, ServerRegistration, ServerRegistrationLen, ServerSetup,
+};
+use rustyline::error::ReadlineError;
+use rustyline::Editor;
+
+// We can define a structure here to hold the parameters for the KSF.
+#[derive(Default)]
+struct CustomKsf {
+    params: scrypt::Params,
+    salt: Vec<u8>,
```
daxpedda, line 33: As far as I understand, adding a salt doesn't make any sense with OPAQUE, it adds no cryptographic value. I think having custom parameters is all the customization one needs here.

@kevinlewi correct me if I'm wrong please.

daxpedda (changes requested):
> 1. I made sure that using `'static` for Argon2 does not break even if you use custom parameters. While you probably still need the `PhantomData` workaround if using a secret key, I can't tell how much of a niche use case that would be (I'm thinking about a case where you'd want to authenticate using both your password and some device key or keyfile).

Great! I'm not entirely sure what the point would be of using a secret key in OPAQUE. If you have a device key or keyfile, you wouldn't need OPAQUE in the first place.
> 2. For the custom `Ksf` example, I copied the entire `simple_login` example and modified it. I also included custom derivation parameters. I'm not sure if that's too overkill, but because of custom parameters we basically need the full login process, so I'm not sure how we could have a shorter "runnable" example.

I don't mind, but any ideas are welcome.
> 3. I think maybe we shouldn't use `Default` for the Ciphersuite because it conflicts with `core::default::Default`. I'd suggest renaming it `DefaultCiphersuite` or something like that while I'm at it.

Yeah, sounds good to me.

examples/custom_ksf.rs
zer0x64: Agreed, the client key used by the server in the OPRF acts as the "salt"/entropy and is basically the way OPAQUE keeps the "salt" secret. I'll remove it!

kevinlewi:
Thanks for putting this up!

I wonder, instead of adding a new runnable example that is mostly a copy of simple_login, could we move the custom KSF example into the docs of lib.rs? I was thinking that we could put it under the "Advanced Usage" section. And the code snippets there could just consist of what you have in the custom_ksf.rs file (the struct definition for `CustomKsf` and the `impl opaque_ke::ksf::Ksf for CustomKsf`).

zer0x64:
> Great! I'm not entirely sure what the point would be of using a secret key in OPAQUE. If you have a device key or keyfile, you wouldn't need OPAQUE in the first place.

It's a somewhat common practice in password managers to use both a password and a keyfile as a form of cryptographically strong 2FA. I know for instance KeePass(XC) allows is and 1Password enforces it.

>Thanks for putting this up!

>I wonder, instead of adding a new runnable example that is mostly a copy of simple_login, could we move the custom KSF example into the docs of lib.rs? I was thinking that we could put it under the "Advanced Usage" section. And the code snippets there could just consist of what you have in the custom_ksf.rs file (the struct definition for CustomKsf and the impl opaque_ke::ksf::Ksf for CustomKsf).

That's a good idea! I did a bit of research and just learned that you can put the `no_run` annotation on docs, which was my worry  since we "should" use a slow hashing function for this kind of example, and passing really weak values wouldn't be a good idea IMO because examples are expected to be copy-pasted.

daxpedda:
> > Great! I'm not entirely sure what the point would be of using a secret key in OPAQUE. If you have a device key or keyfile, you wouldn't need OPAQUE in the first place.
> 
> It's a somewhat common practice in password managers to use both a password and a keyfile as a form of cryptographically strong 2FA. I know for instance KeePass(XC) allows is and 1Password enforces it.

This is a bit off-topic now, but I really enjoy discussing these things :sweat_smile:.
My response was poorly worded, what I wanted to say is that you probably shouldn't use the secret in the KDF of OPAQUE to validate a security key or keyfile. There are other well defined protocols for this which can be run **parallel** to OPAQUE.
> That's a good idea! I did a bit of research and just learned that you can put the `no_run` annotation on docs, which was my worry since we "should" use a slow hashing function for this kind of example, and passing really weak values wouldn't be a good idea IMO because examples are expected to be copy-pasted.

Argon2 is hopefully not bad enough to actually significantly affect our CI times, hopefully `no_run` won't be necessary while still using recommended values.

zer0x64:
So I transferred the examples to the doc. I use `scrypt` to show how to use a foreign KSF and `argon2` to show how to override custom parameters. Since the `scrypt` example does not actually run Scrypt and doesn't require much boilerplate code, I let it live as a doctest. However, I've put the argon2 example as `ignore` for multiple reasons:
- It would need a lot of hidden boilerplate code to get it to compile.
- The `argon2` dependency being optional, I'm not sure how to force it when running the doctest. Without the feature, it simply doesn't compile.
- This example requires a run of Argon2, which as stated because would slow down devellopment cycle because it would run at every `cargo test` command issued.

With that said, I'm happy with the current result, so I'll move the PR out of the draft state!

daxpedda:
> I've put the argon2 example as `ignore` for multiple reasons:

I would strongly discourage from ignoring any documentation test, ultimately it will turn stale and we will have somebody opening an issue saying the code doesn't work.
> * It would need a lot of hidden boilerplate code to get it to compile.

All our documentation tests have a huge amount of boilerplate code.
> * The `argon2` dependency being optional, I'm not sure how to force it when running the doctest. Without the feature, it simply doesn't compile.

You can simply put a `#[cfg(feature = "argon2")]` on top.
> * This example requires a run of Argon2, which as stated because would slow down devellopment cycle because it would run at every `cargo test` command issued.

The current test suit executes a ton of asymmetric cryptography, which is already damn slow, adding a one second Argon2 hashing run is completely insignificant.

> With that said, I'm happy with the current result, so I'll move the PR out of the draft state!

Indeed, looks great!

daxpedda (changes requested):
Other then actually getting the test to run this is great!

kevinlewi (changes requested):
Requesting changes due to failing CI

src/lib.rs
```diff
@@ -986,6 +986,103 @@
...
+//! ```
+//! # use opaque_ke::CipherSuite;
+//! # use opaque_ke::ClientRegistration;
+//! # use opaque_ke::ClientRegistrationFinishParameters;
+//! # use opaque_ke::ServerSetup;
+//! # use opaque_ke::errors::ProtocolError;
+//! # use rand::rngs::OsRng;
+//! # use rand::RngCore;
+//! # use std::default::Default;
+//! #[cfg(feature = "argon2")]
```
daxpedda, line 1028:
```suggestion
//! # #[cfg(feature = "argon2")]
```
That `cfg` can be hidden too.

src/lib.rs
```diff
@@ -986,6 +986,60 @@
 //! let server_setup = ServerSetup::<Default, YourRemoteKey>::new_with_key(&mut OsRng, keypair);
 //! ```
 //!
+//! ## Custom KSF and Parameters
+//!
+//! An application might want to use a custom KSF (Key Stretching Function)
+//! that's not supported directly by this crate. The maintainer of the said KSF
+//! or of the application itself can implement the `Ksf` trait to use it with
+//! `opaque-ke`. `scrypt` is used for this example, but any KSF can be used.
```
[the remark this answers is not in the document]
zer0x64: I reverted the scrypt part since it's an external dev dependencies, which made the CI(and doc formatting) fail

daxpedda (approved):
If the CI is fixed, and except that tiny improvement, LGTM! :tada:

src/lib.rs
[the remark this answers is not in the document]
daxpedda: Yeah, I just saw that. https://github.com/rust-lang/cargo/issues/3475 is what we would need here.

zer0x64:
Both doctests now runs, and the CI passes!

kevinlewi (approved):
This is great, thanks very much for the contribution @zer0x64 !

PR #286, opened 2022-11-22
Upgrade zeroize to 1.5

nitnelave:
This would allow compatibility between this older version and the newer ecosystem.

Alternatively, we could just specify "^1.1" to make sure it's _at least_ 1.1.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -17,7 +17,7 @@ jobs:
           - u32_backend
         toolchain:
           - nightly
-          - 1.41.0
+          - 1.51.0
     name: test
     steps:
       - name: Checkout sources
@@ -94,7 +94,7 @@ jobs:
       matrix:
         toolchain:
           - nightly
-          - 1.41.0
+          - 1.51.0
     name: test simple_login command-line example
     steps:
       - name: install expect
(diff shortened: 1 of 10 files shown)


nitnelave:
I had to push a change with lots of tiny clippy fixes to satisfy the CI, I can split it into a separate PR if you want.

kevinlewi:
The changes look good to me, but out of curiosity, are you hoping to have a new release to the 0.6 branch (0.6.2 for instance)? Or does it suffice for your needs to simply rebase off of the revision containing these patches?

I was not planning on continuing to support the 0.6 branch since it is superseded by v2.0.

kevinlewi (approved):

nitnelave:
If possible, I'd like to have a 0.6.2 release with this patch, but otherwise I can depend on the specific branch.

I know that v2.0 is better, but as we talked about it's data incompatible, so upgrading would require everyone to reset their passwords. I'd like to avoid that if possible.

kevinlewi:
Because this change is not entirely backwards-compatible (increase in MSRV and zeroize version), I don't think we should include it in a 0.6.2. At the very least, it should be a new version (like 0.7), but this might cause confusion with the newer v2.0. Let's keep it without attaching to a specific release for now.

PR #302, opened 2022-12-19
Rename `X25519` to `Curve25519`

### LINKED ISSUE #299: Rename X25519 to Curve25519
I realized that in the past I didn't properly understand the difference between those two words.
The current X25519 implementation doesn't actually use the X25519 key exchange method, it's still using 3DH, so I believe that the appropriate term is Curve25519 instead.

Does that make sense?

daxpedda:
Fixes #299.

### DIFF
--- a/src/key_exchange/group/curve25519.rs
+++ b/src/key_exchange/group/curve25519.rs
@@ -5,7 +5,7 @@
 // License, Version 2.0 found in the LICENSE-APACHE file in the root directory
 // of this source tree.
 
-//! Key Exchange group implementation for X25519
+//! Key Exchange group implementation for Curve25519
 
 use curve25519_dalek::constants::ED25519_BASEPOINT_TABLE;
 use curve25519_dalek::montgomery::MontgomeryPoint;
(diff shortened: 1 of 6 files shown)


CHANGELOG.md
```diff
@@ -84,7 +84,7 @@
 
 * Added CipherSuite API for specifying underlying primitives
 * Added support for specifying a slow password hashing function
-* Collapsed SignalKeyPair to X25519KeyPair
+* Collapsed SignalKeyPair to Curve25519KeyPair
```
kevinlewi, line 87: This can probably not be changed since it is historical.

src/key_exchange/group/curve25519.rs
```diff
@@ -22,11 +22,11 @@ use subtle::ConstantTimeEq;
 use super::KeGroup;
 use crate::errors::InternalError;
 
-/// Implementation for X25519.
-pub struct X25519;
+/// Implementation for Curve25519.
+pub struct Curve25519;
 
 /// The implementation of such a subgroup for Ristretto
```
kevinlewi, line 28: `Ristretto` -> `Curve25519`

kevinlewi (approved):
Two minor nits, but otherwise looks good. Thank you!

kevinlewi (changes requested):
(Setting to request changes)

CHANGELOG.md
daxpedda: Yeah, that was a mistake.

src/key_exchange/group/curve25519.rs
daxpedda: Done.

kevinlewi (approved):
Thanks!

PR #304, opened 2023-01-17
Bump `voprf` to v0.5.0-pre.2

daxpedda:
Depends on https://github.com/facebook/voprf/pull/90 and https://github.com/facebook/voprf/pull/102.
The upgrade to Rust 1.60 also necessitated some Clippy fixes.

PR #306, opened 2023-02-01
Only use explicit crate features

daxpedda:
Analogous to https://github.com/facebook/voprf/pull/100.
Requires https://github.com/facebook/opaque-ke/pull/304.

> Since we are now using Rust 1.60, we can use explicit crate features.
Rust-Analyzer had some serious issues with that `extern crate serde_ as serde`.
This also removes all implicit features, which is an improvement for users.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -12,12 +12,13 @@ rust-version = "1.60"
 version = "2.0.0"
 
 [features]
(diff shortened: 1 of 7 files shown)


daxpedda:
This is ready now!

kevinlewi (approved):
LGTM!

PR #313, opened 2023-03-04
Update `rustyline` to v0.11

daxpedda:
Replaces https://github.com/facebook/opaque-ke/pull/310.


daxpedda:
This increases the MSRV to v1.62, which is w/e really because #307 increases it to v1.65 anyway.

kevinlewi (approved):
Thanks!!

PR #314, opened 2023-03-05
Update `argon2` to v0.5

daxpedda:
Updates `argon2` to v0.5.
Doesn't really include any interesting changes for us except we now don't have to specify our own default salt length, because it's provided by `argon2` now.

See the [CHANGELOG](https://github.com/RustCrypto/password-hashes/blob/argon2-v0.5.0/argon2/CHANGELOG.md#050-2023-03-04).

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -23,7 +23,7 @@ jobs:
           - --features serde
         toolchain:
           - stable
-          - 1.62.0
+          - 1.65.0
     name: test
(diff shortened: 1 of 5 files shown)


daxpedda:
I forgot to mention, this increases the MSRV to v1.62, which is w/e really because #307 increases it to v1.65 anyway.

kevinlewi (approved):
LGTM!

PR #319, opened 2023-03-21
Curve25519 test vectors

daxpedda:
I was trying the test vectors supplied in https://github.com/cfrg/draft-irtf-cfrg-opaque/pull/404 but very quickly ran into a roadblock.

@kevinlewi would appreciate it if you take a look.


src/key_exchange/group/curve25519.rs
```diff
@@ -96,7 +96,7 @@ impl KeGroup for Curve25519 {
         bytes
             .try_into()
             .ok()
-            .and_then(|bytes| Scalar::from_canonical_bytes(bytes).into())
+            .map(Scalar::from_bits_clamped)
```
daxpedda, line 99: For some reason `curve25519-dalek` does not like the keys supplied here. If I understood the `from_bits_clamped()` documentation correctly, it means that the supplied private keys are actually not valid and had to be clamped.

(IANAC and I have no clue what I'm talking about here)

src/tests/test_opaque_vectors.rs
```diff
@@ -238,6 +246,45 @@ fn tests() -> Result<(), ProtocolError> {
...
+        );
+
+        test_registration_request::<Ristretto255Sha512Curve25519NoKsf>(&ristretto_real_tvs)?;
+        test_registration_response::<Ristretto255Sha512Curve25519NoKsf>(&ristretto_real_tvs)?;
+        //test_registration_upload::<Ristretto255Sha512Curve25519NoKsf>(&ristretto_real_tvs)?;
+        //test_ke1::<Ristretto255Sha512Curve25519NoKsf>(&ristretto_real_tvs)?;
+        //test_ke2::<Ristretto255Sha512Curve25519NoKsf>(&ristretto_real_tvs)?;
+        //test_ke3::<Ristretto255Sha512Curve25519NoKsf>(&ristretto_real_tvs)?;
+        //test_server_login_finish::<Ristretto255Sha512Curve25519NoKsf>(&ristretto_real_tvs)?;
+        //test_fake_vectors::<Ristretto255Sha512Curve25519NoKsf>(&ristretto_fake_tvs)?;
```
daxpedda, line 285: These are the tests that aren't working.

src/key_exchange/group/curve25519.rs
daxpedda: Clearly I don't understand what clamping actually means, but for some reason this breaks our current tests which we already generated test vectors for. Meaning clamping here does produce different keys and the keys we were producing until now are wrong?

daxpedda: It does, I was surprised that `curve25519_dalek::Scalar` just is not safe to use then with Curve25519. In any case, I adjusted the code.

I still only got one more test working, the rest fails and I honestly have no clue.

About the test vectors, would it not be better if the test vectors include already pre-clamped private keys? Because I would otherwise say that deserializing private keys should not clamp, it should fail if the key is not valid.

kevinlewi: Yes, let me add a comment to the test vector PR for exactly this. Thanks!

src/tests/test_opaque_vectors.rs
daxpedda: I did try removing the clamping (https://github.com/cfrg/draft-irtf-cfrg-opaque/pull/404#discussion_r1145556262) from the random scalar generation but still couldn't produce the same output.

kevinlewi: A couple of things that I'm noticing (in-progress):
- The generated test vectors still seem to be producing private keys that are not clamped. This is an issue with how they are being generated, and am investigating in the spec PR. Once this is addressed, it should fix the test_ke1 tests.
- We need to adjust the diffie-hellman() function to compute the x25519 function, instead of just multiplying "our" sk with "their" pk. Once this is addressed, it should fix the test_ke2 and test_ke3 tests.

daxpedda: > * We need to adjust the diffie-hellman() function to compute the x25519 function (https://docs.rs/x25519-dalek/latest/x25519_dalek/fn.x25519.html), instead of just multiplying "our" sk with "their" pk.

Unless I'm missing anything the underlying code only does the multiplication too: https://docs.rs/x25519-dalek/2.0.0-pre.1/src/x25519_dalek/x25519.rs.html#317-319.

kevinlewi: Oh right I see, perhaps that part should be fine then :) But yes currently we are waiting on a re-generation of the private keys to ensure that they are clamped.

kevinlewi:
@daxpedda Got the tests working (but am running into issues with pushing and updating this PR). See https://github.com/facebook/opaque-ke/commit/72c473632bcd3b8b85d192bd895aa41daf6f1d85, if you can copy the changes over to this PR here that would be great!

daxpedda:
Amazing!

So there are two points still open:
- Pre-clamping the private keys in the official test vectors. We could then deserializing private keys to fail if they are not valid.
- Figure out if a random oracle is required, which is what we had in place before. See https://github.com/cfrg/draft-irtf-cfrg-opaque/pull/404#discussion_r1148446173.

kevinlewi:
- The pre-clamping in the official test vectors should already be incorporated (see the latest commit in https://github.com/cfrg/draft-irtf-cfrg-opaque/pull/404).
- I don't think a random oracle is required. See the "Computing secret keys" section of https://cr.yp.to/ecdh.html, which I believe is an authoritative source for this kind of stuff.

daxpedda:
> * The pre-clamping in the official test vectors should already be incorporated (see the latest commit in [Add x25519 AKE test vectors cfrg/draft-irtf-cfrg-opaque#404](https://github.com/cfrg/draft-irtf-cfrg-opaque/pull/404)).

I just tested it, doesn't seem to be the case. I want to point out that the latest commit did something with the code but didn't actually update the test vectors, the updated test vectors in the commit are the same as in the commits before, they were just not updated in some of the files. Maybe some oversight happened here.

Just to make sure, maybe there is a misunderstand here, what I mean is that the private keys in the test vectors should already be clamped, instead of having to clamp them during deserialization.
> * I don't think a random oracle is required. See the "Computing secret keys" section of https://cr.yp.to/ecdh.html, which I believe is an authoritative source for this kind of stuff.

I see. 

Out of curiosity, looking at the NIST curves, I couldn't find anything requiring such a random oracle as is currently used in OPAQUE: [NIST FIPS 186-5](https://nvlpubs.nist.gov/nistpubs/FIPS/NIST.FIPS.186-5.pdf#%5B%7B%22num%22%3A156%2C%22gen%22%3A0%7D%2C%7B%22name%22%3A%22XYZ%22%7D%2C70%2C409%2C0%5D).

Probably better to ask this in the IETF OPAQUE repo, but I couldn't really find an explanation why we need such a "heavy-handed" random oracle for P-256/P-384/P-521.

kevinlewi:
> I just tested it, doesn't seem to be the case. I want to point out that the latest commit did something with the code but didn't actually update the test vectors, the updated test vectors in the commit are the same as in the commits before, they were just not updated in some of the files. Maybe some oversight happened there.
> 
> Just to make sure, maybe there is a misunderstanding here, what I mean is that the private keys in the test vectors should already be clamped, instead of having to clamp them during deserialization.

Ah, oops! I found an issue with the official test vectors which I thought were clamping properly... but looks like there is still a bug. Sorry, will get that sorted out! But yes, we are on the same page with what needs to happen :)

daxpedda:
Updated the test vectors.
Apparently `randomized_pwd` is missing on some, but otherwise looks good!

daxpedda:
Alright, everything works now!
Thanks you @kevinlewi for the amazing work!

kevinlewi:
FYI, I think the test vectors had some variable names changed, in case this causes errors:
server_keyshare -> server_public_keyshare
client_keyshare -> client_public_keyshare
randomized_pwd -> randomized_password

daxpedda:
Adjusted.
`client_keyshare` was removed from the test vectors in the latest commit of https://github.com/cfrg/draft-irtf-cfrg-opaque/pull/404, I assume this was by accident?

kevinlewi:
> `client_keyshare` was completely removed from the test vectors in the latest commit of [cfrg/draft-irtf-cfrg-opaque#404](https://github.com/cfrg/draft-irtf-cfrg-opaque/pull/404), it was not replaced with `client_public_keyshare`, I assume this was by accident?

Ah, seems like `client_keyshare` / `client_public_keyshare` was mistakenly removed in https://github.com/cfrg/draft-irtf-cfrg-opaque/pull/407. I will push another commit to add it back in, thanks for the catch!

daxpedda:
Alright, this is good to go now!

kevinlewi:
This is in sync with https://github.com/cfrg/draft-irtf-cfrg-opaque/pull/404/commits/a55d914a5207c6490de54c14cf6be59bf7f8bfd3, but after https://github.com/cfrg/draft-irtf-cfrg-opaque/pull/411 there were a couple of more minor-ish edits to the test vectors which need to be incorporated. We can do so in a follow-up PR, as I'd like to merge this one first.

PR #321, opened 2023-05-13
Fix zeroize dependency in a more future-proof fashion

nitnelave:
Also, bump to 0.6.2 for release.

I'm trying to publish a crate that depends on opaque 0.6 and (indirectly) on zeroize 1.5.7. The 0.6.1 of opaque_ke is still on zeroize ~1.1, so it's not compatible.

For me to release my crate, I'd need a 0.6.2 release.

(note that I _know_ that 0.6 is deprecated, but my crate will depend on _both_ opaque 0.6 and 3.0+ and will allow users to do a smooth transitions without having to reset their passwords)

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -31,7 +31,7 @@ rand = "0.8"
 serde = { version = "1", features = ["derive"], optional = true }
 subtle = { version = "2.3.0", default-features = false }
 thiserror = "1.0.22"
-zeroize = { version = "~1.5", features = ["zeroize_derive"] }
+zeroize = { version = "1.5", features = ["zeroize_derive"] }
 
 [dev-dependencies]
 anyhow = "1.0.35"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "opaque-ke"
-version = "0.6.1"
+version = "0.7.0"
 repository = "https://github.com/novifinancial/opaque-ke"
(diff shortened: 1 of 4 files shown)


daxpedda:
@kevinlewi pointed out here (https://github.com/facebook/opaque-ke/pull/286#issuecomment-1327832553) why a v0.6.2 won't be released with a higher version of `zeroize`: the MSRV.

Maybe a 0.7 version would be more appropriate?

Cargo.toml
```diff
@@ -31,7 +31,7 @@ rand = "0.8"
 serde = { version = "1", features = ["derive"], optional = true }
 subtle = { version = "2.3.0", default-features = false }
 thiserror = "1.0.22"
-zeroize = { version = "~1.5", features = ["zeroize_derive"] }
+zeroize = { version = "^1.5", features = ["zeroize_derive"] }
```
daxpedda, line 34: That's the same as no qualifier:
```suggestion
zeroize = { version = "1.5", features = ["zeroize_derive"] }
```

nitnelave:
Good point. Changed to 0.7

kevinlewi:
LGTM. I updated the MSRV in the CI to 1.56. I can commit this, but @nitnelave can you test on the updated commit before I make an actual 0.7.0 release?

nitnelave:
Yep, just built and tested with the new commit, it works for me. Thanks!

PR #323, opened 2023-05-21
Correctly clamp Curve25519 secret keys

daxpedda:
Before we release the next version with #322, we should probably backport the bugs we found in https://github.com/facebook/opaque-ke/pull/319.

I will rebase https://github.com/facebook/opaque-ke/pull/319 afterwards.


kevinlewi (approved):
Thanks for breaking this out!

PR #326, opened 2023-05-26
Update keypair generation to use derive_auth_keypair

kevinlewi:
In https://github.com/cfrg/draft-irtf-cfrg-opaque/pull/411, the test vectors were updated to ensure that keypair generation was done using the specified derivation method (using DeriveDiffieHellmanKeyPair). The test vectors also were updated to include the seed used to generate the keypair, as opposed to the private key itself. The constant string "OPAQUE-DeriveAuthKeyPair" was also updated to now be "OPAQUE-DeriveDiffieHellmanKeyPair"

Previously, this implementation just generated the private key randomly and derived the corresponding public key in a way that matched the test vectors. Now that the test vectors have become more strict, it revealed that we were not generating them according to specification.

This PR fixes that, and also pulls in the new test vectors to make sure they match.


kevinlewi:
cc: @daxpedda

src/keypair.rs
```diff
@@ -58,8 +63,20 @@ impl<KG: KeGroup, S: SecretKey<KG>> KeyPair<KG, S> {
...
+    where
+        <CS::Hash as OutputSizeUser>::OutputSize:
+            IsLess<U256> + IsLessOrEqual<<CS::Hash as BlockSizeUser>::BlockSize>,
+    {
+        let mut scalar_bytes = GenericArray::<_, <KG as KeGroup>::SkLen>::default();
+        rng.fill_bytes(&mut scalar_bytes);
+        let sk = KG::derive_auth_keypair::<CS>(
+            GenericArray::clone_from_slice(&scalar_bytes),
+            &STR_OPAQUE_DERIVE_AUTH_KEY_PAIR,
+        )
```
daxpedda, line 78: I believe we can just remove the `info` parameter from `derive_auth_keypair` and hardcode it into the method.

daxpedda (commented):
LGTM!

PR #328, opened 2023-06-08
Fixing simple_login test to enable argon2 feature

kevinlewi:


daxpedda:
Strange that it worked until now, according to Cargo this should have been necessary all along.
There is even an issue to change that behavior: https://github.com/rust-lang/cargo/issues/4663.

Probably not worth digging into.

PR #329, opened 2023-06-15
update parameter from sk to private_key

nikgraf:


daxpedda (commented):
It's not a bad idea, user facing API should probably not use `sk` naming.
I approve!

kevinlewi (approved):
Thanks!

PR #330, opened 2023-06-24
Bump `curve25519-dalek` to v4.0.0-rc.3

daxpedda:
The `Scalar` type isn't compatible with X25519 anymore, so instead we use a plain `[u8; 32]`.

Requires https://github.com/facebook/voprf/pull/113.


daxpedda:
This is ready now.

PR #342, opened 2023-09-22
Updating workflow dependencies

kevinlewi:
Also increases MSRV to 1.70


daxpedda:
Increasing the MSRV should probably also be noted in the changelog.

Issue #343, opened 2023-09-26
Cannot get connection successful after server restart


DorianCoding:
Hello and sorry if the question is evident.

I have implemented opaque and it works well when the server is up. However if I restart the server, I cannot connect anymore.

I have the same private key from ServerSetup using new_from_key and the public key is printed and is the same but the OsRng is different. Is it the problem ? if yes, how to save the OsRng parameters?

Of course, the password file is read succesfully each time.

Client :
```
target/debug/client -r
Username: dorian
Password:

Please confirm:

Connecting to remote host...
Register mode
Connected to [::1]:5501
Sending HELLO
Received HELLO, authentification phase.
Authentification calculation begin...
▰▰▰▰▰▰▰ Done                                                                                                                                                                                            

Registered

$ target/debug/client
Username: dorian
Password:

Connecting to remote host...
Login mode
Connected to [::1]:5502
Sending HELLO
Got hello, authentification phase
Got challenge
▰▰▰▰▰▰▰ Done                                                                                                                                                                                            

Shared Key is K******* and my key is 8**********

$ target/debug/client
Username: dorian
Password:

Connecting to remote host...
Login mode
Connected to [::1]:5502
Sending HELLO
Got hello, authentification phase
Got challenge
▰▰▰▰▰▰▰ Done                                                                                                                                                                                            

Error during calculation, definitely wrong password.
```
Server (your public key represents the key from ServerSetup) :
```
Your public key is: jhdHAQAVZNlJe-dkslkTg6OzWq5e6XIzNwtTeyZX8gs
Listening sockets
The username hash : 'p__G-L8e12ZRwUdWoGHWYvWA_03kO0n6gtgKS4D4Q0o' is not present, sending dummy login.
Error: No result, probably incorrect password. from [::1]:37692
Connection accepted with [::1]:36522.
Registration successful for user 'dorian'
VALID: Connection from [::1]:58732 - KEY: K********
^C
$ target/debug/server
Your public key is: jhdHAQAVZNlJe-dkslkTg6OzWq5e6XIzNwtTeyZX8gs
Listening sockets
Error: No result, probably incorrect password. from [::1]:51460
```
The so-called code : 
```
let mut rng = OsRng;
let server_setupraw: ServerSetup<DefaultCipherSuite> = match fs::read_to_string(keyfile) {
        Ok(key) => {
            let mut keybuffer: &mut [u8] = &mut [0; 500];
[...]

PR #344, opened 2023-09-28
Clarifying the persisting of server setup

kevinlewi:
Addresses https://github.com/facebook/opaque-ke/issues/343

### DIFF
--- a/src/opaque.rs
+++ b/src/opaque.rs
@@ -224,6 +224,10 @@ where
     Le<<<OprfHash<CS> as CoreProxy>::Core as BlockSizeUser>::BlockSize, U256>: NonZero,
 {
     /// Create [`ServerSetup`] with the given keypair
+    ///
+    /// This function should not be used to restore a previously-existing
+    /// instance of [ServerSetup]. Instead, use [ServerSetup::serialize] and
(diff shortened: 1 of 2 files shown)


src/opaque.rs
```diff
@@ -224,6 +224,10 @@ where
     Le<<<OprfHash<CS> as CoreProxy>::Core as BlockSizeUser>::BlockSize, U256>: NonZero,
 {
     /// Create [`ServerSetup`] with the given keypair
+    ///
+    /// This function should not be used to restore a previously-existing
+    /// instance of [ServerSetup]. Instead, use [ServerSetup::serialize] and
+    /// [ServerSetup::deserialize] for this purpose.
```
daxpedda, line 230:
```suggestion
    /// instance of [`ServerSetup`]. Instead, use [`ServerSetup::serialize`] and
    /// [`ServerSetup::deserialize`] for this purpose.
```

kevinlewi: The styling for the rest of the documentation is not using backticks for the structs / functions at the moment, but I'll try changing this style for the entire file to see how it looks and will address these changes in its own PR.

PR #290, opened 2022-12-09
Test P-384

daxpedda:
This just tests P-384 support and builds on top of https://github.com/novifinancial/voprf/pull/84, which builds on top of https://github.com/RustCrypto/elliptic-curves/pull/600.
We still have to wait for new versions to be released.

~~This also builds on top of https://github.com/novifinancial/opaque-ke/pull/288, which builds on top of https://github.com/novifinancial/opaque-ke/pull/289.~~
Builds on top of https://github.com/facebook/opaque-ke/pull/307.

PR #349, opened 2023-11-10
Test P-521

daxpedda:
This adds testing of P-521 through the [`p521`](https://crates.io/crates/p521) crate.

Still requires:
- [x] https://github.com/RustCrypto/elliptic-curves/pull/964
- [x] https://github.com/RustCrypto/elliptic-curves/issues/965

See #290 for a previously similar addition.
See https://github.com/facebook/voprf/pull/127 for the same PR in `voprf`, which isn't required to merge this.


daxpedda:
CI fails because of https://github.com/RustCrypto/elliptic-curves/issues/965, which we are going to wait for anyway.

daxpedda:
This is ready to merge now.

kevinlewi:
It seems like the maintenance for full_test.rs is getting more and more cumbersome now... any thoughts for how we can make this simpler?

And maybe for now, can you also add (in a comment above each new set of constants in full_test.rs) instructions for the cargo command to re-generate those constants?

daxpedda:
Last two commits should address both concerns.
WDYT?

kevinlewi (approved):
Oh wow, this is really awesome -- thank you for that refactor! Looks good to me

PR #358, opened 2024-08-19
docs: add details for client login final step

lucab:
This tweaks the documentation on the main module, in order to add some details on the outcome of the client login final step. In particular, it clarifies the result of `ClientLogin::finish()` both on success and on errors and it adds some intra-crate links to the relevant structures and fields.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -390,11 +390,18 @@
 //!
 //! ### Client Login Finish
 //! In the third step of login, the client takes as input a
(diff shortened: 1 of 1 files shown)


lucab:
@kevinlewi thanks for maintaining this library!
I was recently learning about the protocol and after reading the crate docs I was left with some minor questions about the client-side login finish step. I hope this small PR will make it easier for other newcomers like me to navigate through the docs and the login flow.

PR #365, opened 2024-09-21
chore!: align to draft 16, update dependencies

cyyynthia:
This PR updates the implementation to (actually) match the draft 16 of the protocol. Most notable (and breaking) change is the change of the context string to `OPAQUEv1-`. Terminology around aPAKE has been updated to match the changes done in the draft as well (from asymmetric to augmented).

I've also tweaked the cargo deny config to get it working on newer versions (as well as bumping the action to v2), and ditched unmaintained json dependency in favor of serde_json.

Side note, that is probably more of a protocol-level nit than something to do with the lib: [RFC 9497](https://www.rfc-editor.org/rfc/rfc9497#section-3.1) (OPRF) uses an uppercase V1, whereas [OPAQUE](https://datatracker.ietf.org/doc/html/draft-irtf-cfrg-opaque#section-6.4.2.1) uses a lowercase v. Very specific, but an inconsistency nonetheless 😄


kevinlewi:
Ah, thank you, this is awesome!!

And nice catch on my mistake of claiming it was updated to version 16 without actually having updated it in the previous release. Apologies on that.

kevinlewi (approved):

PR #370, opened 2025-04-15
Bump MSRV to v1.83

daxpedda:
- Update MSRV to v1.83. Now a lot of our trait constraints aren't viral anymore, but are implied! This gets rid of all the `where` constraints on structs and implementations!
- With the new MSRV, we can now implement `Error` for our error types without `std` as well.
- Updated `rustyline`, a dev-dependency.
- Updated `unicode-ident`s license exception from `Unicode-DFS-2016` to `Unicode-3.0`.

This is a huge improvement to the documentation as well, which was littered with these huge `where`-clauses.

In the same spirit as https://github.com/facebook/voprf/pull/140.


kevinlewi (approved):
LGTM!

PR #372, opened 2025-04-19
Create dummy `masking_key` in `ServerSetup`

daxpedda:
Currently we generate a new masking key every time we create a dummy `ServerRegistration` in response to a non-existing user. This isn't ideal because using the random generator might change the time it takes to respond to the user request making it possible for an adversary to determine if a user record exists or not.

This new implementation is not constant time, because we are still creating a `ServerRegistration` ad-hoc, but there is nothing involved apart from stack memory operations, which should not have any impact compared to querying a database for an existing user record.

We could make this constant-time by storing a dummy `ServerRegistration` in `ServerSetup` and implement [`ConditionallySelectable`](https://docs.rs/subtle/latest/subtle/trait.ConditionallySelectable.html) for `ServerRegistration`. But I believe its unnecessary.

This change is motivated by the [following paragraph in the specification](https://www.ietf.org/archive/id/draft-irtf-cfrg-opaque-18.html#section-6.3.2.2-7):
> It is RECOMMENDED that a fake client record is created once (e.g. as the first user record of the application) and then stored alongside legitimate client records, to serve subsequent client requests. This allows servers to retrieve the record in a time comparable to that of a legitimate client record.

(storing alongside legitimate client records doesn't make much sense to me, because most databases don't have a constant-time implementation of "if you can't find this specific record return me this specific (dummy) record")


daxpedda:
After thinking about this a bit more, it seems to have been a good idea to me to re-generate the masking key on demand. This prevents a leakage of the dummy masking key to compromise the servers ability to protect against user enumeration.

I will create a new PR changing the current setup to constant-time then and open an issue at the draft repo.

PR #373, opened 2025-04-19
Remote OPRF Seed Support

daxpedda:
This PR expands the API to enable storing the OPRF seed on an HSM. The operation that is externalized here is the HDKF expand operation in `CreateRegistrationResponse` and `CreateCredentialResponse`, which is used to derive the per-user OPRF key. The API and documentation call the output of the HKDF operation "key material".

The change is rather small and the API flow only changes minimally. Instead of calling `ServerRegistration::start()`, users have to call `ServerSetup::key_material_info()` first, generate the key material and then call `ServerRegistration::start_with_key_material()`. The same with `ServerLogin::builder()`, users generate the key material and then call `ServerLogin::builder_with_key_material()`.

Notably the current API does not allow to have a remote OPRF seed but a local private key. My assumption is that it is unlikely anybody would run a setup like that.

I should also mention that I'm not aware of any HSM supporting HKDF, not even SoftHSM. PKCS#11 itself only received support in v3 AFAIK. Instead users are expected to implement HKDF by hand and use the HSMs HMAC operation, which almost all HSMs out there support. This still keeps the OPRF seed protected.

Changes:
- `ServerSetup::new_with_key_pair_and_seed()` to build a `SeverSetup` from a remote private key *and* OPRF seed.
- `ServerSetup::key_material_info()` returns a `KeyMaterialInfo`, containing all necessary information to generate the key material needed via HKDF.
- `OprfSeedSerialization`, the equivalent of `PrivateKeySerialization` too facilitate `ServerSetup::de/serialize()` with a remote OPRF seed.
- `ServerRegistration::start_with_key_material()`, the equivalent of `ServerRegistration::start()` but with a remote OPRF seed.
- `ServerLogin::builder_with_key_material()`, the equivalent of `ServerLogin::builder()` but with a remote OPRF seed.
- Adjusted the remote key test to make use of the new functionality.

Based on #371.

### DIFF
--- a/src/keypair.rs
+++ b/src/keypair.rs
@@ -23,20 +23,20 @@ use crate::key_exchange::tripledh::DiffieHellman;
     feature = "serde",
     derive(serde::Deserialize, serde::Serialize),
     serde(bound(
-        deserialize = "S: serde::Deserialize<'de>",
-        serialize = "S: serde::Serialize"
+        deserialize = "SK: serde::Deserialize<'de>",
+        serialize = "SK: serde::Serialize"
     ))
 )]
 #[derive_where(Clone)]
-#[derive_where(Debug, Eq, Hash, Ord, PartialEq, PartialOrd; KG::Pk, S)]
-pub struct KeyPair<KG: KeGroup, S: Clone = PrivateKey<KG>> {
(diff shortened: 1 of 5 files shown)


kevinlewi (approved):
OK, this looks good. Thanks!

PR #374, opened 2025-04-22
Only Store Dummy Public Key

daxpedda:
This is split off from #372. For the dummy registration file we only need the public key.


kevinlewi (approved):
Thanks!

daxpedda:
We shouldn't merge this before #378, I would like to avoid having to rebase it.
Putting it in draft mode.

PR #376, opened 2025-04-22
Always create dummy record

daxpedda:
This change makes sure the dummy record is always created, even if a user record is available. This prevents any timing difference caused by operating the RNG when a record is not available.

Follow-up from #372.


daxpedda:
Created https://github.com/cfrg/draft-irtf-cfrg-opaque/issues/474.

kevinlewi (approved):
Looks good, thanks!

By the way, the OPAQUE draft is currently in the RFC Editor's queue, which means that the time for edits other than from the RFC Editor has already passed...

daxpedda:
> Looks good, thanks!
> 
> By the way, the OPAQUE draft is currently in the RFC Editor's queue, which means that the time for edits other than from the RFC Editor has already passed...

I see, that's unfortunate.

kevinlewi:
@daxpedda: By the way, do you mind adding the copyright header to `src/tests/remote_key.rs`, either in this PR or the next one? Unfortunately I missed the omission of this in #371. Thanks in advance!

daxpedda:
Done in https://github.com/facebook/opaque-ke/pull/377.

PR #378, opened 2025-05-04
SIGMA-I Key Exchange

daxpedda:
I again apologize beforehand for the (almost) useless commit history. They contain a lot of iterations that are just overwritten later and improvements I've just run into on the way. I'm happy to extract any unrelated improvements into their own PR on demand.

---

# Motivation

The main reason is HSM support. It has been more than 8 years now since Cloud HSMs have become readily accessible, but unfortunately their list of available algorithm is still quite limited. Specifically DH is currently, among popular Cloud HSMs, [only available on AWS](https://docs.aws.amazon.com/kms/latest/APIReference/API_DeriveSharedSecret.html).

However, because its main use-case is TLS all Cloud HSMs support signing.

Another motivation is PQC. SIGMA-I could easily be used with ML-DSA/SLH-DSA and ML-KEM. While, as [the specification says](https://www.ietf.org/archive/id/draft-irtf-cfrg-opaque-18.html#appendix-C-2), it wouldn't make OPAQUE quantum-safe, because the OPRF isn't, it would at least provide forward secrecy (?) for the session key.

<h1 name="sigma-i-implementation">Implementation</h1>

While [the specification lists it under alternative key exchanges](https://www.ietf.org/archive/id/draft-irtf-cfrg-opaque-18.html#name-sigma-i-instantiation-sketc), unfortunately the description is incomplete and partly outdated.

The last more comprehensive description of SIGMA-I in OPAQUE was in [the draft version v1](https://www.ietf.org/archive/id/draft-irtf-cfrg-opaque-01.html#name-instantiation-with-sigma-i), so I ended up following that.

Here is a complete list of differences between the SIGMA-I and 3DH AKE protocol implementations in OPAQUE as implemented by this PR:
- The [`ikm` passed to `DeriveKeys`](https://www.ietf.org/archive/id/draft-irtf-cfrg-opaque-18.html#section-6.4.2.2-2) only consists of one shared secret, a DH between the server and client ephemeral keys, instead of three shared secrets.
- The [`server_mac` only takes the `server_identity`](https://www.ietf.org/archive/id/draft-irtf-cfrg-opaque-18.html#section-6.4.4-2) instead of a hash of the `preamble`. The same applies to the [`expected_server_mac`](https://www.ietf.org/archive/id/draft-irtf-cfrg-opaque-18.html#section-6.4.3-4).
- The [`expected_client_mac` only takes the `client_identity`](https://www.ietf.org/archive/id/draft-irtf-cfrg-opaque-18.html#section-6.4.4-2) instead of a hash of the `preamble` + `server_mac`. The same applies to the [`client_mac`](https://www.ietf.org/archive/id/draft-irtf-cfrg-opaque-18.html#section-6.4.3-4).
- The server additionally sends a signature signed by the servers static private key of the transcript *without* `client_identity` in [`AuthResponse`](https://www.ietf.org/archive/id/draft-irtf-cfrg-opaque-18.html#section-6.1-8).
- In [`AuthClientFinalize`](https://www.ietf.org/archive/id/draft-irtf-cfrg-opaque-18.html#section-6.4.3-4) the client verifies the signature sent in `AuthResponse`.
- Then the client signs the transcript *without* `server_identity` and sends it in [`KE3`](https://www.ietf.org/archive/id/draft-irtf-cfrg-opaque-18.html#section-6.1-15).
- In [`AuthServerFinalize`](https://www.ietf.org/archive/id/draft-irtf-cfrg-opaque-18.html#section-6.4.4-4) the server verifies the signature sent in `KE3`.

This outline makes sense to me according to [the SIGMA-I paper](https://www.iacr.org/cryptodb/archive/2003/CRYPTO/1495/1495.pdf#5.2) as well. The only difference being that we sign much more than outlined in the paper, but the MAC seems about right. IANA a cryptographer, so I would really appreciate some feedback on this.

<h1 name="notable">Notable Changes</h1>

## Cipher Suite

The [`KeGroup`](https://github.com/facebook/opaque-ke/blob/b16cf8ce05f868325c780f00aaba1d19cf70a175/src/ciphersuite.rs#L38-L39) was removed from `CipherSuite` and the VOPRF hash is now separated from the [`KeyExchange`](https://github.com/facebook/opaque-ke/blob/b16cf8ce05f868325c780f00aaba1d19cf70a175/src/ciphersuite.rs#L41). I talk more about that in [Key Exchange](#key-exchange).

<h2 name="key-exchange">Key Exchange</h2>

The [`KeyExchange`](https://github.com/facebook/opaque-ke/blob/b16cf8ce05f868325c780f00aaba1d19cf70a175/src/key_exchange/traits.rs#L22) trait had to be modified extensively to make this work. However, many of these changes improve the API and also fix some bugs.

<h3 name="separate-hash">Separate Hash</h3>

Until now the [`KeyExchange`](https://github.com/facebook/opaque-ke/blob/b16cf8ce05f868325c780f00aaba1d19cf70a175/src/key_exchange/traits.rs#L22) trait had two generics: a group and a hash. The [`CipherSuite` trait sets both of these generics to `CipherSuite::KeGroup` and `voprf::CipherSuite::Hash`](https://github.com/facebook/opaque-ke/blob/b16cf8ce05f868325c780f00aaba1d19cf70a175/src/ciphersuite.rs#L41). But the OPRF and key exchange actually don't interact with each other and therefor the key exchange hash should not depend on the OPRF.

While this is a minor inconvenience for the key exchange, it is wrong for `KeGroup::derive_auth_keypair()`. The `hashToScalar()` implementation uses the OPRF hash and `DeriveAuthKeyPair()` uses that with a DST containing the ID from the OPRF cipher suite. So we mix the OPRF ID and hash with the key exchange curve and its `DeriveAuthKeyPair()` algorithm.

This isn't just an issue of not following the spec, this becomes a security issue if for example the OPRF suite is using P-256 with SHA-256 and the key exchange is using P-521. [The specification demands that the hash must at least match the curves target security](https://www.ietf.org/archive/id/draft-irtf-cfrg-opaque-18.html#section-7-2.2.1), which would require SHA-512 here. However, I doubt that in practice anybody chose a configuration like this.

To that end, both generics of the [`KeyExchange`](https://github.com/facebook/opaque-ke/blob/b16cf8ce05f868325c780f00aaba1d19cf70a175/src/key_exchange/traits.rs#L22) trait are removed. Instead, `KeGroup::derive_auth_keypair()` now uses [`voprf::derive_key()`](https://docs.rs/voprf/latest/voprf/fn.derive_key.html) directly. The key exchange implementation, e.g. [`TripleDh`](https://github.com/facebook/opaque-ke/blob/b16cf8ce05f868325c780f00aaba1d19cf70a175/src/key_exchange/tripledh.rs#L61), gains two generics: a group and a hash.

So an old `CipherSuite` implementation would look like this:
```rust
impl CipherSuite for DefaultCipherSuite {
    type OprfCs = Ristretto255;
    type KeGroup = Ristretto255;
    type KeyExchange = TripleDh;
    type Ksf = Identity;
}
```
The new one looks like this:
```rust
impl CipherSuite for DefaultCipherSuite {
    type OprfCs = Ristretto255;
    type KeyExchange = TripleDh<Ristretto255, Sha512>;
    type Ksf = Identity;
}
```

Now the key exchange implementation is in full control of its own cipher suite. However, the OPAQUE implementation still needs to know which kind of types its dealing with, so `KeyExchange` gains two associated types: `Group` and `Hash`. For `TripleDh` this now looks like this:
```rust
impl<G: Group, H: Hash> KeyExchange for TripleDh<G, H> {
    type Group = G;
    type Hash = H;
    ...
}
```

Again, this shows that the `KeyExchange` trait gains more control, deciding which types exactly go into e.g. [`ServerSetup`s `KeyPair`](https://github.com/facebook/opaque-ke/blob/b16cf8ce05f868325c780f00aaba1d19cf70a175/src/opaque.rs#L71). To illustrate the point, we can take a look at the new SIGMA-I key exchange:
```rust
impl<SIG: SignatureProtocol, KE: Group, KEH: Hash> KeyExchange for SigmaI<SIG, KE, KEH> {
    type Group = SIG::Group;
    type Hash = KEH;
    ...
}

impl<G: Group, H: Hash> SignatureProtocol for Ecdsa<G, H> {
    type Group = G;
    ...
}

impl CipherSuite for DefaultCipherSuite {
    type OprfCs = Ristretto255;
    type KeyExchange = SigmaI<Ecdsa<Nist521, Sha512>, Ristretto255, Sha512>;
    type Ksf = Identity;
}
```

Now we have two groups and two hashes to deal with and the implementation is able to determine which one goes where.

#### Migration

This will cause issues with migration for users who have so far used two different curves for the OPRF and key exchange. The only way users can fix this is by implementing their own custom `Group`, copying our implementation, but changing `Group::derive_scalar()` to keep the old implementation.

I can add a small migration guide in the root documentation if its desirable.

### SIGMA-I

A new type [`SigmaI`](https://github.com/facebook/opaque-ke/blob/1e325fbf0ec1335f1118a0dfc88d178597273bea/src/key_exchange/sigma_i/mod.rs#L68) is introduced, which implements the SIGMA-I key exchange protocol as [outlined above](#sigma-i-implementation).

This introduces a bunch of things, but the important part is the [`SignatureProtocol` trait](https://github.com/facebook/opaque-ke/blob/1e325fbf0ec1335f1118a0dfc88d178597273bea/src/key_exchange/sigma_i/mod.rs#L86). This allows for different implementation to add their own types, generics and implementation more independently. E.g. currently ECDSA takes a generic hash, while EdDSA does not. I predict ML-DSA and SLH-DSA will also appreciate this level of control. Another example is PureEdDSA and HashEdDSA. Both can now be composed with different groups. E.g.:
```rust
type KeyExchange = SigmaI<PureEddsa<Ed25519>, ...>;
type KeyExchange = SigmaI<PureEddsa<Ed448>, ...>;
type KeyExchange = SigmaI<HashEddsa<Ed25519>, ...>;
type KeyExchange = SigmaI<HashEddsa<Ed448>, ...>;
```
... instead of:
```rust
type KeyExchange = SigmaI<Ed25519, ...>;
type KeyExchange = SigmaI<Ed448, ...>;
type KeyExchange = SigmaI<Ed25519Ph, ...>;
type KeyExchange = SigmaI<Ed448Ph, ...>;
```

#### ECDSA

As already seen earlier, a [new type `Ecdsa` is introduced](https://github.com/facebook/opaque-ke/blob/1e325fbf0ec1335f1118a0dfc88d178597273bea/src/key_exchange/sigma_i/ecdsa.rs#L39), which takes a `elliptic-curve` group and a hash and implements ECDSA for it.

Currently only `NistP256` and `NistP384` are supported, because `NistP521` doesn't use the same shared types as the others. However, this has been resolved in the newer versions of `elliptic-curve` which are still in beta.

The implementation doesn't use the high-level trait offered by [`ecdsa`](https://crates.io/crates/ecdsa) because generating `k` is fixed to the curves assigned hash, even if the message is hashed with a different one. See https://github.com/RustCrypto/signatures/issues/949. So we are doing the implementation with [`ecdsa`s low-level functions](https://docs.rs/ecdsa/0.16.9/ecdsa/hazmat/index.html). This comes with the advantage of users not having to enable the `ecdsa` crate feature of their curves crate, which in turn doesn't pull in the default hash's crate.

<h4 name="eddsa">EdDSA</h4>

EdDSA is implemented as well, in two forms: [`PureEddsa`](https://github.com/facebook/opaque-ke/blob/1e325fbf0ec1335f1118a0dfc88d178597273bea/src/key_exchange/sigma_i/pure_eddsa.rs#L29) and [`HashEddsa`](https://github.com/facebook/opaque-ke/blob/1e325fbf0ec1335f1118a0dfc88d178597273bea/src/key_exchange/sigma_i/hash_eddsa.rs#L28).

The HashEdDSA implementation is straightforward as it works like ECDSA and can simply pre-hash the client message to verify the signature upon receiving it. However PureEdDSA requires the whole message during verification and our current implementation was using a bunch of slice references, which couldn't be stored in `Ke2State` without introducing lifetimes in the high-level API of the crate and all the complexity this comes with. This lead to a large refactor of all the `KeyExchange` parameter types [described below](#ke-refactor).

Another issue with PureEdDSA is that `ed25519-dalek` doesn't support message streaming, see https://github.com/dalek-cryptography/curve25519-dalek/pull/556. Our message can't be put together into a single slice because we don't want to allocate. `concat()`ing it together isn't an option either because it includes slices provided by the user with an unknown size at compile time. So I ended up copying the implementation code from `ed25519-dalek` to support this.

Note that our implementation doesn't not violate the specifications ["do not implement a IUF interface"](https://www.rfc-editor.org/rfc/rfc8032#section-8.7). On the input side we have the message prepared, just not in a single slice of bytes. On the output side we don't continue the protocol until the signature has been *completely and successfully* verified.

PureEdDSA will be especially important for users who want to use EdDSA over ECDSA, because most popular Cloud HSMs only support PureEdDSA, e.g. [Google Cloud KMS](https://cloud.google.com/kms/docs/algorithms#elliptic_curve_signing_algorithms).

I briefly looked at implementing CtxEdDSA (not a formal name) while I was at it. But unfortunately OPAQUE allows a `context` size of over 255 bytes, so it doesn't really map with Ed25519/448ctx's limits.

Currently only [`ed25519-dalek`](https://crates.io/crates/ed25519-dalek) is supported. In the future we want to add support for Ed448 via https://github.com/RustCrypto/elliptic-curves/pull/1121.

### Removed Serialization bounds

The various key exchange states and messages were required to implement [`Serialize` and `Deserialize`](https://github.com/facebook/opaque-ke/blob/b16cf8ce05f868325c780f00aaba1d19cf70a175/src/key_exchange/traits.rs#L84-L92). This required [a lot of where bounds on `TripleDh`](https://github.com/facebook/opaque-ke/blob/b16cf8ce05f868325c780f00aaba1d19cf70a175/src/key_exchange/tripledh.rs#L184-L197) and when [using `TripleDh` bounds elsewhere](https://github.com/facebook/opaque-ke/blob/b16cf8ce05f868325c780f00aaba1d19cf70a175/src/tests/full_test.rs#L1513-L1534).

Now they are not required and instead de/serialization implementations that require it can simply [add a single where bound on those](https://github.com/facebook/opaque-ke/blob/1e325fbf0ec1335f1118a0dfc88d178597273bea/src/opaque.rs#L619).

### Publicly Accessible

I moved the [`remote_key` test](https://github.com/facebook/opaque-ke/blob/b16cf8ce05f868325c780f00aaba1d19cf70a175/src/tests/remote_key.rs) to be an [integration test](https://doc.rust-lang.org/1.86.0/cargo/reference/cargo-targets.html#integration-tests) to see what kind of access it needed to work properly. I found out it was quite tough to get things working without access to the `KeyExchange` trait.

So now the [`KeyExchange`](https://github.com/facebook/opaque-ke/blob/b16cf8ce05f868325c780f00aaba1d19cf70a175/src/key_exchange/traits.rs#L22) trait itself is now publicly accessible. However, it is [sealed](https://github.com/facebook/opaque-ke/blob/1e325fbf0ec1335f1118a0dfc88d178597273bea/src/key_exchange/traits.rs#L36) and can't be implemented by dependents. Furthermore, only the new `type Group` and `type Hash` are visible, [all other associated types and methods are `doc(hidden)`](https://github.com/facebook/opaque-ke/blob/1e325fbf0ec1335f1118a0dfc88d178597273bea/src/key_exchange/traits.rs#L47-L64).

I want to note here that while I think we are getting closer to publicly expose the trait fully, it seems that as of yet nobody has even requested that.

<h3 name="ke-refactor">(Internal) Refactored Parameter Types</h3>

As mentioned in [the EdDSA section](#eddsa), SIGMA-I needs to verify the client signature when reaching server login finish. Which is a problem for PureEdDSA because it doesn't use pre-hashes and needs the actual message during the signature verification process.

This lead me to cleaning up and encapsulating a lot of the parameters passed into the `KeyExchange` trait methods. While there is still more work to be done, I think we are a big step closer towards exposing the `KeyExchange` trait publicly.

## High-Level API Changes

`ClientLogin` now takes an RNG and `ServerLogin` takes a `ServerLoginStartParameters`, which is now renamed to `ServerLoginParameters`.

At first I tried quite hard to save the context and identifiers in `KeState` somehow, but introducing these lifetimes made the API quite horrible.

<h2 name="de-serialization">(Internal) De-serialization Overhaul</h2>

Until now de-serialization code was [incredibly complex and error prone](https://github.com/facebook/opaque-ke/blob/b16cf8ce05f868325c780f00aaba1d19cf70a175/src/messages.rs#L442-L483). By implementing a [mutable reference to a reference slice take method](https://github.com/facebook/opaque-ke/blob/1e325fbf0ec1335f1118a0dfc88d178597273bea/src/serialization/mod.rs#L79-L94), we can [very easily and safely de-serialize complex types](https://github.com/facebook/opaque-ke/blob/1e325fbf0ec1335f1118a0dfc88d178597273bea/src/messages.rs#L403-L418) without manually having to check for correct sizes for every type.

# Changelog

## Notable

See the [Notable Changes](#notable) section for a more detailed explanation.

- Remove `CipherSuite::KeGroup` in favor of associated type `type Group` on `trait KeyExchange`.
- Add `type Group` and `type Hash` associated types to `trait KeyExchange`.
- Add group and hash generics to `trait TripleDh`.
- Add `trait SigmaI`, implementing the SIGMA-I key exchange for OPAQUE.
- Add `Ecdsa` implementing ECDSA for `SigmaI`. It is gated behind the `ecdsa` crate feature and depends on [`ecdsa`](https://crates.io/crates/ecdsa).
- Add `Ed25519` implementing Ed25519 as a signature for `SigmaI`. It is gated behind the `ed25519` crate feature and depends on [`ed25519-dalek`](https://crates.io/crates/ed25519-dalek).
- Remove the `ristretto255-voprf` crate feature. `ristretto255`s `derive_scalar()` now directly relies on its `voprf::CipherSuite`.
- Partly publicly expose `trait KeyExchange`. Only `type Group` and `type Hash` are visible however, all other associated types and methods are `doc(hidden)`.
- `ClientLogin::finish()` now requires passing a RNG.
- `ServerLogin::finish()` now requires `ServerLoginStartParameters`.
- Rename `ServerLoginStartParameters` to `ServerLoginParameters`, as it is used in both start and finish.
- Remove default `Group::hash_to_scalar()` and let current implementations call `voprf` directly. See [Separate Hash](#separate-hash) section for more details on how this fixes a bug. This also removes `Group::is_zero_scalar()`, which was only used in the default implementation of `Group::hash_to_scalar()`.
- Replace `impl Iterator` and `&[u8]` parameter types in `trait KeyExchange` methods with owned types. `CredentialRequestParts`, `CredentialResponseParts`, `SerializedIdentifiers` and `SerializedContext`. This encapsulates a lot of code in `opaque.rs` and therefor cleaned it up a lot.

## Minor

- Remove `De/Serialize` requirements on `trait KeyExchange` associated types. These bounds were not able to be implied by Rust and [forced usage of e.g. `TripleDh` to list all bounds](https://github.com/facebook/opaque-ke/blob/b16cf8ce05f868325c780f00aaba1d19cf70a175/src/tests/full_test.rs#L1613-L1638). Instead, users can just [bind on `De/Serialize` directly](https://github.com/facebook/opaque-ke/blob/1e325fbf0ec1335f1118a0dfc88d178597273bea/src/tests/full_test.rs#L1352) (which are currently not publicly exposed!).
- Switch `elliptic-curve` implementation to use [`NonZeroScalar`](https://docs.rs/elliptic-curve/0.13.8/elliptic_curve/scalar/struct.NonZeroScalar.html) and [`NonIdentity`](https://docs.rs/elliptic-curve/0.13.8/elliptic_curve/point/struct.NonIdentity.html). This was important to avoid panicking code in the implementation.
- Rename `Group::derive_auth_keypair()` to `Group::derive_scalar()`. This name was coming from an older draft where the function was called `GenerateAuthKeyPair()`. But the current draft (v18) has two functions, `DeriveKeyPair()` and `DeriveDiffieHellmanKeyPair()`. Our method covers both and only creates the scalar, so it doesn't map exactly to the how its presented in the specification.
- Add [`MaskedResponseLen` bounds to `CipherSuite`](https://github.com/facebook/opaque-ke/blob/1e325fbf0ec1335f1118a0dfc88d178597273bea/src/ciphersuite.rs#L37-L42), remove the need have the bounds on [`ClientLogin::finish()`](https://github.com/facebook/opaque-ke/blob/b16cf8ce05f868325c780f00aaba1d19cf70a175/src/opaque.rs#L491-L495) and [`ServerLogin::start()`](https://github.com/facebook/opaque-ke/blob/main/src/opaque.rs#L723-L727)
- `KeyExchange::KE2Builder` now carries a lifetime.
- Remove [`ExpandMsg` bounds on `CipherSuite` for OPRFs hash](https://github.com/facebook/opaque-ke/blob/b16cf8ce05f868325c780f00aaba1d19cf70a175/src/ciphersuite.rs#L29-L30). `Group` implementation which need it now call `voprf` directly.

## Testing

- Exclude the `src/test/` folder from Cargo publishing. I usually do the opposite and only include files relevant for dependents, but I will leave this discussion to a separate PR. It was important here because `full_test_vectors.rs` has reached 0.6 MB already.
- Introduce `trait AssertZeroized`. This is needed to allow tests to continue functioning which for all the new types that don't `zeroize()` to zero, like `NonZeroScalar`. However, all of our `Zeroize` tests can be removed now, we use `derive_where` and have no custom `Zeroize` implementations left. But I will leave this to another PR.
- Move the `remote_key` test to an integration test, to make sure all relevant types are publicly accessible.
- Generate ciphersuites in the `full_test_vectors` test via a macro, covering all possible combinations, especially with the new SIGMA-I addition.

### Notable

- Rework de-serialization implementations. See [Deserialization Overhaul](#de-serialization) section. Rename all *internal* `deserialize(_*)` functions to `deserialize_take(_*)`.
- Rename `KeGroup` to `Group`. Now that there is a clear separation between OPRF and the key exchange, it seemed unnecessary and odd to me to try and differentiate it from the `voprf::Group` trait.
- Publicly expose `OprfSeedSerialization` in `opaque_ke::keypair`.

### Minor

- Rename all uses of a generic `KG` for the key exchange group to `G`. This is to avoid confusion because now key exchange implementations can have multiple groups.
- Rename all uses of a generic `D` for hash to `H`.
- Add `OprfSeed`, a container for the OPRF seed implementing `OprfSeedSerialization`. This is to avoid implementing `OprfSeedSerialization` over [`Zeroizing`](https://docs.rs/zeroize/1.8.1/zeroize/struct.Zeroizing.html).
- Add code example for remote OPRF seed in the root documentation.
- Slightly improve formatting in the root documentation code examples by adding some newlines and better control visible imports.
- Split `KeyPair::generate_random()` into `KeyPair::random()` and `KeyPair::derive_random()`. `KeyPair::random()` is used in `ServerSetup`, which does not use hash2curve. `KeyPair::derive_random()`, which as before relies on `Group::derive_scalar()` is used in generating key shares.
- Remove `ProtocolError::IdentityGroupElementError`. This was used to check for identity elements during de-serialization of OPRF types, however, this is already done during `voprf` de-serialization.
- Move client mac calculation on the server side to server login start. This makes `Ke2State` a bit lighter and also maps with [`expected_client_mac` in the specification](https://www.ietf.org/archive/id/draft-irtf-cfrg-opaque-18.html#section-6.4.4-2).
- Remove rejection sampling for `Curve25519::random_sk()` and other unnecessary zero scalar checks. `clamp_integer()` already accounts for that. I added a unit test as well just to make sure.
- Add internal `PrivateKey::new()` constructor to avoid having to serialize a `Group::Sk` and then use `PrivateKey::deserialize()` to get a new `PrivateKey`.

### Testing

- Fix `cross-test` in CI accidentally using crate features as test filters and not actually running any tests.
- Add non-HSM custom OPRF seed test.

---

There are a couple of things I wanna follow up on here in future PRs:
- Finish cleaning up `trait KeyExchange` so we can make it publicly accessible.
- Remove more serialization bounds by binding them on `CipherSuite`.
- See if we can simplify our current `hash::Hash` trait.
- Document all error cases on every function.
- Overhaul all our tests to make them less verbose. They are currently quite repetitive because we keep adding more and more possible cipher suites we want to test.


kevinlewi (approved):
Ok, this is a huge set of changes but looks good! Thanks for the clear documentation overall.

I think it would be helpful to add a caveat to the doc comments of mod.rs of sigma_i, something like:

```
//! ⚠️ **Warning**: This implementation has not been audited. Use at your own risk!
```

Otherwise, happy to land the changes.

daxpedda:
Done.

src/key_exchange/sigma_i/message.rs
```diff
@@ -0,0 +1,296 @@
+use core::ops::Add;
```
kevinlewi, line 1: This file is missing the license header

kevinlewi:
Thanks! I will go ahead and merge this in first since it's quite a large change, but if you could address the missing header from `sigma_i/message.rs` in a follow-up, that would be great!

PR #380, opened 2025-05-05
Move Serde De/Serialization to `Group` Implementation

### LINKED ISSUE #300: Use `Public/SecretKey` for the `elliptic-curve` implementation
Currently we are using `ProjectivePoint` and `Scalar` directly instead of `PublicKey` and `SecretKey` respectively.

The reason was that `curve25519_dalek::Scalar` supports `Zeroize` but not `ZeroizeOnDrop`.
But `elliptic_curve::SecretKey` supports `ZeroizeOnDrop` but not `Zeroize`.
(this was one of the main reasons I worked on introducing `ZeroizeOnDrop` in the first place)

The best solution here would be to introduce a `SecretKey` for Ristretto255, which currently doesn't exist, we could introduce this type ourselves in `opaque-ke` though.

Not sure what to do about `elliptic_curve::PublicKey`, as it implements neither. Is not zeroing out the public key acceptable?

This is analogous to https://github.com/novifinancial/voprf/issues/57 and is mainly for code improvement and cleaner deserialization, it doesn't actually change anything.

daxpedda:
Until now we had a manual Serde De/Serialization implementation for `Public/PrivateKey`. This PR changes that to shift the responsibility on the `Group` implementation. Now `Group` implementations can choose their own way to format their keys.

As `elliptic-curve` is the only implementation with dedicated types for non-identity elements and non-zero scalars, all other ones now require custom Serde implementations.

Resolves #300.
Based on #378.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -24,6 +24,7 @@ serde = [
   "curve25519-dalek?/serde",
   "ecdsa?/serde",
   "ed25519-dalek?/serde",
+  "elliptic-curve/serde",
   "generic-array/serde",
   "voprf/serde",
   "zeroize/serde",
--- a/src/key_exchange/group/curve25519.rs
+++ b/src/key_exchange/group/curve25519.rs
@@ -80,6 +80,7 @@ impl Group for Curve25519 {
 }
 
 /// Curve25519 scalar.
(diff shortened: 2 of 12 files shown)


kevinlewi:
Can you rebase + rerun the automated checks before I merge these changes in? Thanks!

daxpedda:
PR up in #383.

PR #381, opened 2025-05-06
Remove Zeroize Tests

daxpedda:
We originally introduced them because we had a macro generating the `Zeroize` implementations before we switched to `derive_where`. In addition, zeroizing became more nuanced because the types we are depending on don't zeroize to 0, but to e.g. 1. So we ended up writing manual code that checks if something has been zeroized according to its own rules. Writing tests that test code only written for tests is a bit of nonsense.

So this PR removes that. We also remove the `Zeroize` requirements on `Group::Pk`, as discussed in #300.

Based on #378.
Based on #380.


kevinlewi (approved):
Looks good. Thank you!

PR #386, opened 2025-09-08
Updating test vectors to RFC 9807

kevinlewi:
No protocol changes, just updating the test vectors to match what is in RFC 9807. Also updating the changelog to prepare for a new version release.


daxpedda (commented):
Do we want to mention somewhere how to upgrade existing `ServerSetup`s and `RegistrationUpload`s?

We could simply add some migration code right into the changelog. Or add it to the root documentation.

Let me know and I'm happy to make a PR.

